Issue #10, opened 2021-02-02
Import all contracts data and ABI from github


AugustoL:
Right now we are compiling the smart contracts from solidity files, BUT we already have a repository with all the dxdao contracts that needs to be used, we want to import the contracts from there https://github.com/AugustoL/dxdao-contracts

PR #40, opened 2021-04-17
Add hardhat

AugustoL:
- Remove contracts folder and import contracts using hardhat dependency compiler from dxdao-contracts repository.
- Update deploy local contracts script to work with hardhat.
- Remove truffle config file.
- Add hardhat config file.


AugustoL:
fix #10

PR #140, opened 2021-09-12
Added boolean check to convert from string to boolean

rossneilson:
Currently boolean values get interpreted as strings and so getting left blank = false and any text ("true" / "false") ends up getting encoded as true.
To fix this I added an additional check like we do for arrays to convert the text to a real boolean (Line 321).

Another thing noticed during testing is that the edits we do to arrays and now booleans gets reflected in the UI causing doubts about whether something has gone wrong. I will make an issue (#141) to prevent changes in the UI related to encoding done away from the user. 

Also I might be going mad but I am fairly sure my IDE is using the prettier file but all files I change get formatted. Is this my problem or is the code just not been formatted in a while?

### DIFF
--- a/src/pages/NewProposal.tsx
+++ b/src/pages/NewProposal.tsx
@@ -50,7 +50,7 @@ const TitleInput = styled.div`
       border: 1px solid gray;
       padding: 0px 5px;
     }
-    
+
     select {
       margin-left: 5px;
       background-color: white;
@@ -69,7 +69,7 @@ const TextActions = styled.div`
     flex-direction: column;
     margin: 10px 0px;
     line-height: 30px;
-    
+
 `;
 
 const CallRow = styled.div`
@@ -78,7 +78,7 @@ const CallRow = styled.div`
     justify-content: left;
     flex-direction: row;
     margin-bottom: 10px;
-    
+
     span {
       text-align: center;
       font-family: Roboto;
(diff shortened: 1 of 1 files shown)


src/pages/NewProposal.tsx
```diff
@@ -15,780 +15,998 @@ import {
...
-              return callData;
+
+            if (call.dataValues.length > 0) {
+              call.functionParams.map((functionParam, i) => {
+                if (functionParam.type.indexOf('[]') > 0) {
+                  call.dataValues[i] = call.dataValues[i]
+                    .slice(1, -1)
+                    .split(',');
+                }
+                if (functionParam.type === 'bool') {
```
AugustoL: This is the only piece of code you added?

rossneilson: Yes

AugustoL:
@rossneilson about the code format change, yes, it is not formatted but we have rules now.

I would suggest you if you can only submit the piece of code you change, you can do it by doing a reset soft (remove commits but not changes done in code) and then only commit what you changed in the dapp to fix the issue. 

Later if you can do another commit with the changes to update the source code to folloow the styling rules.

`git reset HEAD~1 --soft ` Uncommit latest commit but keep your changes

`git add -p` go file but file and slect only the piece of code you want to commit

`git commit` or I use `git cz` an alias to git commitizen to have interactively do standarized commits https://commitizen-tools.github.io/commitizen/

rossneilson:
Ok, I messed up the commits a lot there and just reset everything. My gitfu was not strong enough :D
Even with formatting on save disabled VS code removed the empty lines but at least the first commit is now cleaner.
I'll redo the formatting now also so it is in another commit (#142)

AugustoL (approved):

PR #145, opened 2021-09-14
Add WalletConnect support.

### LINKED ISSUE #51: Add WalletConnect back

### LINKED ISSUE #81: Update Web3React and add more connectors
We should do an update/refactor of the connectors used, follow the code exposed in https://github.com/NoahZinsmeister/web3-react.

Maybe the best way would be to update to something even simpler like https://github.com/web3modal/web3modal

Ideally we want more connectors beside metamask. Simplify and a bit the code and update it.

mprasanjith:
- [x] Add WalletConnect connector and update wallet switcher UI.
- [x] Add support for using custom RPC endpoints with WalletConnect.
- [x] Set custom Infura RPC API Key.
- [x] Set custom Alchemy RPC API Key and let user choose between Alchemy and Infura.
- [x] Let user override RPC URL completely.
- [X] Hide the custom RPC configs when using injected / MetaMask wallet.

Closes #51 and also partially fixes #81.

**Known Issue:**
We do not have a way of switching between networks if the user is connected via WalletConnect. We should implement a network switcher in DXvote for this.


AugustoL (approved):

AugustoL:
It would be good to figure out in a future a stable RPC that can be used by default, maybe pokt network, that is used by one.

thx @mprasanjith great contribution

PR #148, opened 2021-09-14
Update github workflows, issues templates and contributing.

AugustoL:


rossneilson (approved):
We should add IPFS CI build at some point to get rid of netlify

AugustoL:
> We should add IPFS CI build at some point to get rid of netlify

The good thing about netlify is that it give us a developer url, the same for github.io in master branch.

I added an IPFS script that hash the build, so later we can verify it with our local guilds for official releases.

I would rather improve the IPFS CI script to leave somehow a tag o link in the readme with the hash of the release build (built and hashed using CI)

The idea would be in official releases run the IPFS hash script, and check it againts our local builds and thats it, that would be the hash of the official release.

PR #153, opened 2021-09-16
Caching data using Cache API

### LINKED ISSUE #80: Save cache data in local storage.
Each time the dapp loads it gtes the data from the cache block to the last block in the network to update the dao cache to the latest block.

Save the data fetched in the local storage, this way the data for the dapp would be `dxvote cache + localstorage cache + latest data`.

Importat: Dont save the data of the dapp in the cache, it is already available in the dapp, the goal here is to reduce the loading time if you enter to the app often.

rossneilson:
Initially the intended solution was to use local storage which was implemented in the first commit however there is a limit of 5mb here for the domain and so mainnet for example could not be cached. 
The solution found to this was to use the browser cache API which is meant for use by service workers and has no size limit. 
https://developer.mozilla.org/en-US/docs/Web/API/Cache
This is supported on all browsers other than IE. I'm confident it would still load using IE just without caching although it would be good to have some way of testing this since I am on mac (not that anyone should be using IE anyways and I dont think it is a priority for us to support it).

There was another issue found where when the from and to block for updating were equal the recent updated block value was being updated but since the to block number was not higher than from, no data was actually fecthed.
This wasn't a big issue previously as it would be fetched again on reload, however with caching this doesnt happen. So now this should be fixed and I confirmed that REP was updating properly when redeemed.

I also added the delete function to the cache clear button in settings.

Closes #80


src/stores/BlockchainStore.ts
```diff
@@ -166,33 +166,43 @@ export default class BlockchainStore {
...
+        const cache = await caches.open(`dxvote-cache`);
+        let match = await cache.match(networkName);
+        let networkCache = null;
+        if (match) {
+          networkCache = JSON.parse(await match.text());
+        }
+
+        const blockNumber = (await library.eth.getBlockNumber()) - 1;
+
+        // Fetch cache from ipfs if not in localStorage or > ~4 days old
```
AugustoL: Or if the hash of the cache saved in storage is different to the one hosted online.You will have to also store the IPFS hash of the cache stored.

Bacause if you get the cache now with hash A and it changes today to hash B it will only be fetched 3 days from today, right? So first lets check the hash of the cache stored and then time.

AugustoL (changes requested):

src/stores/BlockchainStore.ts
rossneilson: Oh smart! So I can remove the 4 days check, store the hash and check it agains the most recent one for whether it needs fetched or not

AugustoL (approved):

PR #158, opened 2021-09-17
Default should not show unboosted proposals

### LINKED ISSUE #147: Default should not show unboosted proposals
**Is your feature request related to a problem? Please describe.**
When looking at the proposal view in DXvote, it sorts by relevant proposals, which is mostly based upon when the proposal is finished. This makes sense for boosted proposal but it gets clogged up when there are unboosted (often mistake) proposals that are expiring. As can be seen here.
Screen Shot 2021-09-14 at 3 39 21 PM

The Amphiboly proposals are not relevant. Luzzifoss is

**Describe the solution you'd like**
I think the default view should only show 'boosted proposals' or maybe if it has a certain amount of REP voted for it (for the case of majority votes?)

hhamud:
- Closes #147 
- Default sorting now priorities boosted proposals and In-Queue proposals. 
- Example
image


rossneilson:
Some merge issues here

hhamud:
> Some merge issues here

what do you mean?. the utils/cache file?

rossneilson:
Yeah it has the merge <<<<<develop things, looks like a merge gone wrong
Oh they are being removed

rossneilson (approved):

PR #160, opened 2021-09-19
Improve load times

rossneilson:
It took a bit of debugging but I managed to find what was slowing down the UI, I'll explain what I did here.

So previously we had load times of over 20 seconds, with recent improvements from Augusto and the caching I implemented this was improved a little however even in a best case caching scenario load times were still over 15 seconds long.

There was no network activity and placing log messages all over I could not find what was taking so long to load before the actual complex task of fetching data begun. 

The performance tab of chrome developer tools however showed a huge idle time in between processes with absolutely nothing running:
スクリーンショット 2021-09-18 18 34 36

I eventually found the issue in the interval for fetching data, we set the interval but didn't call it immediately, this drastically improved load times. I also found when first setting the interval there is an if statement that is false and so doesn't run immediately but only runs again once the delay passes and so I made the delay dynamic fixing the problem.

With a fresh cache locally load times are now in the 3-4 second range.

From here some more optimisations may be possible but likely the most effective method will be reducing bundle size.

### DIFF
--- a/src/stores/BlockchainStore.ts
+++ b/src/stores/BlockchainStore.ts
@@ -186,7 +186,6 @@ export default class BlockchainStore {
             await ipfsService.getContentFromIPFS(newestCacheIpfsHash)
           );
           networkCache.baseCacheIpfsHash = newestCacheIpfsHash;
-          daoStore.setCache(networkCache);
         }
 
         const lastCheckedBlockNumber = networkCache.l1BlockNumber;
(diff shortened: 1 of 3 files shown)


src/stores/BlockchainStore.ts
```diff
@@ -256,12 +255,12 @@ export default class BlockchainStore {
           networkCache.l1BlockNumber = toBlock;
           providerStore.setCurrentBlockNumber(toBlock);
 
-          daoStore.setCache(networkCache);
           await cache.put(
             networkName,
             new Response(JSON.stringify(networkCache))
           );
         }
+        daoStore.setCache(networkCache);
```
rossneilson, line 263: This was a bug I missed when implementing caching

AugustoL (approved):

PR #163, opened 2021-09-20
Feature/modal confirmation

### LINKED ISSUE #137: Confirm your vote pop-up
When voting on DXvote, you click on the thumbs up or thumbs down icon and then it immediately prompts the wallet to confirm and send the transaction. It's incredibly important to make sure you are voting for what you intend for.

I like how on Alchemy it gives you a run down of what the vote will be _after_ your vote
Screen Shot 2021-09-10 at 1 28 55 PM

rossneilson:
Added a confirmation popup before voting on a proposal as we have in alchemy.
I was going to add a pie chart style component also but I will create another issue for this later.

Previously we were using a package for modals however it was not adding much other than an extra dependency bloating our bundle size so I refactored it to use our own. Part of the code was taken from the existing modal in Aqua, thanks @mprasanjith 
Tested with existing wallet modal to ensure exact same styles and behaviour.

Closes #137 

スクリーンショット 2021-09-20 14 21 14
スクリーンショット 2021-09-20 14 21 19


src/pages/Proposal.tsx
```diff
@@ -643,11 +653,8 @@ const ProposalPage = observer(() => {
               })}
           </span>
           <span style={{ width: '50%', textAlign: 'center', color: 'red' }}>
-            {proposal.negativeVotes
-              .times('100')
-              .div(totalRepAtProposalCreation)
-              .toFixed(2)}{' '}
-            %<AmountBadge color="red">{negativeVotesCount}</AmountBadge>
+            {`${positiveVotes}%`}
```
AugustoL: this should be negativeVotes, found this bug when I went to vote in xdai and I saw that the  negative voting amout suddenly changed after I click on voting

AugustoL (changes requested):

src/pages/Proposal.tsx
rossneilson: Oops, fixed now

AugustoL (approved):

PR #168, opened 2021-09-22
Take in count proposingRepReward for Redeems 

### LINKED ISSUE #131: Take in count proposingRepReward for Redeems
Right now in dxvote if you create a proposal in a scheme that is giving REP to the proposal creator and you dont stake or vote in the proposal you wont be able to redeem since we are not taking in count that case in the redeem calculation.

The best way to solve it would be to get the proposals created by the user and compare it against the redeems rep events. (in the cases where the proposal was created in a scheme where the `proposingRepReward` is higher than 0)

hhamud:
* Closes #131 
* Adds in a new mapping over the `newProposal` creations in `userEvents`
* refactored one of the if statements to make it more readable.
* Removed commited out code in `proposal.tsx `
* Added new utils `hasLostReputation` , `isBoosted`, `isExpired`, `isNotActive`,`isWinningVote,`

### DIFF
--- a/src/stores/DaoStore.ts
+++ b/src/stores/DaoStore.ts
@@ -848,6 +848,21 @@ export default class DaoStore {
       bounty: [],
     };
 
+    // Adds user created proposals that have ended
+    userEvents.newProposal.map(newProposal => {
+      const proposal = this.getProposal(newProposal.proposalId);
+      const votingParameters = this.getVotingParametersOfProposal(
+        newProposal.proposalId
+      );
+
+      if (
+        votingParameters.proposingRepReward.toNumber() > 0 &&
+        isNotActive(proposal)
+      ) {
+        redeemsLeft.rep.push(newProposal.proposalId);
+      }
+    });
+
     // Add possible redeems
     userEvents.votes.map(vote => {
       const proposal = this.getProposal(vote.proposalId);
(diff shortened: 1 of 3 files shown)


rossneilson (approved):
All looks good to me, nice job adding those util methods !

src/utils/proposals.ts
```diff
@@ -5,6 +5,36 @@ import {
 } from './index';
 import moment from 'moment';
 
+export const isExpired = (proposal: Proposal): boolean => {
+  return (
+    proposal.stateInVotingMachine === VotingMachineProposalState.ExpiredInQueue
+  );
+};
+
+export const isBoosted = (proposal: Proposal, vote: Vote): boolean => {
```
AugustoL, line 14: this can be voteBeforeBoosted and only return is the vote was done before boosted, thats what we care about.
You can also remove the notBoosted constant, not needed since you already have a boolean value on boosted

src/stores/DaoStore.ts
```diff
@@ -843,38 +848,54 @@ export default class DaoStore {
...
-            vote.timestamp < proposal.boostedPhaseTime.toNumber()) ||
-            proposal.boostedPhaseTime.toNumber() === 0)) ||
-          (voteParameters.votersReputationLossRatio.toNumber() > 0 &&
-            ((proposal.boostedPhaseTime.toNumber() > 0 &&
-              vote.timestamp < proposal.boostedPhaseTime.toNumber()) ||
-              proposal.boostedPhaseTime.toNumber() === 0) &&
-            proposal.winningVote === vote.vote &&
-            proposal.stateInVotingMachine < 3)) &&
-        redeemsLeft.rep.indexOf(vote.proposalId) < 0
+        (isExpired(proposal) && isBoosted(proposal, vote)) ||
```
AugustoL, line 874: Awesome that you refactored this, I think we shoudl check it like this, before the proposal has to be expired, and if we have rep locked higher than zero and the vote was done before boosted(if it did boosted, a proposal can execte without reaching boosted) and if the vote was on the winning option
```
isExpired(proposal)
  && (hasLostReputation(voteParameters) && voteBeforeBoosted(proposal, vote))
  && isWinningVote(proposal, vote)
  && redeemsLeft.rep.indexOf(vote.proposalId) < 0)
```

PR #176, opened 2021-09-24
Proposals view style changes

rossneilson:
Initially this was some style changes to the existing view for proposals however I went a bit far and changed the overall layout to what I think is a more modern look.

If people dont like the design I can revert it, I'm aware we should have designers doing this and have a mockup made beforehand but making this in figma would take longer than just doing it so ...

I decided to do this after seeing how commonwealth designed their page which has a similar purpose to ours but looked a lot better.

スクリーンショット 2021-09-27 16 46 08

### DIFF
--- a/src/components/Footer/index.tsx
+++ b/src/components/Footer/index.tsx
@@ -15,6 +15,7 @@ const LeftFooter = styled.div`
   margin-top: auto;
   display: flex;
   flex-direction: row;
+  flex-wrap: wrap;
 `;
 
 const FooterItem = styled.div`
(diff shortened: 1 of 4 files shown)


AugustoL:
image
The left sidebar looks very weird IMO.

I would recommend having the same width in all left sidebar buttons, set to 100% in the sidebar and with centered text.
This might even allow us increase the width of the proposals section, so the sidebar can be a fixed width and the proposals section taje the remaining width.

I would also either add a background to the sidebar or a table border to the proposals to separate the both components, I think a table border with a change in row color on mouse hover would be the best option.

rossneilson:
Thanks for the suggestions! will do that, and made a few other changes also I will push shortly

AugustoL (approved):
Technically approved but I added my comments on what I should think we should change in the ui.

rossneilson:
Updated now!

PR #184, opened 2021-09-27
Proposal sorting

### LINKED ISSUE #179: Proposal sorting
**Description**
For the proposal view, imo the default order of proposals as you scroll should be:
Boosted
Pre-boosted
executed succeeded 
In queue 

perhaps, you could switch 'executed succeeded' and 'in queue', but in the current form all of the 'in queue' proposals are before the current boosting proposals.
Screen Shot 2021-09-24 at 12 20 54 PM
Screen Shot 2021-09-24 at 12 20 43 PM

The current boosting proposals are *VERY IMPORTANT*. A lot of ways the boosting period in holographic consensus could be more important than the voting period (think back to contentious proposals, and it was on the boosting phase). It should be easy to see 'what is boosting' right now. Perhaps that's actually completely separate from the voting experience? Or it's a different user really.

hhamud:
* Refactored the `sortedProposal` variable
* Created a helper `MapEnum` function
* Created a new check that sorts  `boostTime` first if that exists
* Closes #179

### DIFF
--- a/src/pages/Proposals.tsx
+++ b/src/pages/Proposals.tsx
@@ -150,15 +150,11 @@ const ProposalsPage = observer(() => {
 
   const sortedProposals = allProposals
     .filter(proposal => proposal.stateInVotingMachine > 4)
-    .sort(function (a, b) {
-      return a.finishTime - b.finishTime;
-    })
+    .sort((a, b) => a.finishTime - b.finishTime)
     .concat(
       allProposals
         .filter(proposal => proposal.stateInVotingMachine < 5)
-        .sort(function (a, b) {
-          return b.finishTime - a.finishTime;
-        })
+        .sort( (a, b) => b.finishTime - a.finishTime)
     );
 
   if (sortedProposals.length > proposals.length) setProposals(sortedProposals);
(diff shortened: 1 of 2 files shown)


src/utils/helpers.ts
```diff
@@ -2,6 +2,19 @@ import { utils } from 'ethers';
...
 import { MAX_UINT } from './constants';
 
+type EnumType = { [s: number]: string };
+
+export function mapEnum(enumerable: EnumType, fn: Function): any[] {
+  // get all the members of the enum
+  let enumMembers: any[] = Object.keys(enumerable).map(key => enumerable[key]);
+
+  // we are only interested in the numeric identifiers as these represent the values
+  let enumValues: number[] = enumMembers.filter(v => typeof v === 'number');
```
rossneilson, line 13: Bit confused why we need to filter numbers out, is there anything else in there?

src/pages/Proposals.tsx
```diff
@@ -148,19 +150,22 @@ const ProposalsPage = observer(() => {
   // lower in the finish time.
   // This way we show the proposals that will finish soon first and the latest proposals that finished later
```
AugustoL: We should change the comment that describes how it is ordered, right?

src/utils/helpers.ts
hhamud: updated the functions comments to clear this up.

AugustoL (approved):

PR #199, opened 2021-09-29
Feature/162 blockchain link

luchux:
Done:
	•	added utils methods for: a) getENSName(address) b) getERC20token(address) c) getDxVoteContract(address)
	•	refactored repeated implementation of addressStub function and move to helpers.
	•	implemented an async call in mount for the component to call helper for check for ENS name resolution.
WIP: 
refactor the way the BlockchainLink presents the info in the component such that:
-  if a) show ens name with link to address explorer, and an icon to copy address.

 -if b) show token token symbol with link to token explorer, and icon to copy address. if c) show contract name with link to address explorer, and icon to copy address.
 
Rethink the way this component is used to define the correct props regarding use cases in the page components.


src/utils/address.ts
```diff
@@ -21,8 +22,80 @@ export function toChecksum(address) {
...
+    default: // investigate DRY here
+      if (networkName === 'arbitrum')
+        return `https://explorer.arbitrum.io/#/tx/${address}`;
+      else if (networkName === 'arbitrumTestnet')
+        return `https://rinkeby-explorer.arbitrum.io/#/tx/${address}`;
+      else if (networkName === 'mainnet')
+        return `https://etherscan.io/tx/${address}`;
+      else if (networkName === 'xdai')
+        return `https://blockscout.com/xdai/mainnet/tx/${address}`;
+      else return `https://${networkName}.etherscan.io/tx/${address}`;
```
hhamud: I think it's best to expand the parameters of the supported chain in the constants.ts file. 

```typescript

export const SUPPORTED_CHAINS: {
[NETWORK_IDS.arbitrum]: {
explorer: `...`,
rpc: `...`
   }
}
```
and then to use that function to alter the URLs, this will make it easier to maintain in the future.

src/utils/address.ts
```diff
@@ -21,8 +22,80 @@ export function toChecksum(address) {
...
+  }
+}
+
+export async function getENSName(address) {
+  let name = null;
+  try {
+    const provider = ethers.getDefaultProvider();
+    let checksumed = ethers.utils.getAddress(address);
```
hhamud: I don't understand why this is a variable and not just a constant?.

hhamud (changes requested):
Good overall changes.

src/utils/address.ts
luchux: oh yeah, good call! I disliked this spagethi but didn't face the refactor ;)

src/utils/address.ts
luchux: uf, bad on me.

luchux:
@rossneilson I have merged mprasanjith work  (still draft) on my branch  to reuse the constants definition of networks and some of the getter functions. 
When his PR is merged, safe to merge this. Meanwhile worth having a look at this PR ;) 

Thanks! 
BTW: given that there is not test engine as to write some tests (to see that everything renders as expected in each address case), what would be the ideal way of editing some of the hardcoded proposals to add addresses that are in the tokens, and dx contracts ? (for users, ens and addresses with the loaded data is enough).

rossneilson:
I havent really looked through the code yet for review but just a few pointers from the preview build.
There are still a lot of locations where we arent showing the ENS name instead of address. 
For example, proposal history, the top right for connected user.
I also wonder if it's worth adding a shortener to the ENS names also since they cause some weird formatting issues when longer ones are displayed.

luchux:
@rossneilson @AugustoL 
thinks to resolve are: 
- Where do we take the contracts from dxVote. In the config we have all networks tokens and contracts, but I think we look contracts in dxVote according to this issue specifications. which are those? 
- styling issues (showing some screenshots) (I don't want to take my own decision here, since seems we look for consensus)

This is a dxDao token found on the config for all this addresses (a test in where all the addresses are a token address) 
Screen Shot 2021-10-08 at 11 51 30 AM

This is an AAVE token found on the config for all this addresses (a test in where all the addresses are a token address) 
Screen Shot 2021-10-08 at 12 41 34 PM

This are addresses resolved to an ENS (a test in where each address resolves to smolpp.eth)
Screen Shot 2021-10-08 at 11 51 45 AM

AugustoL:
> @rossneilson @AugustoL thinks to resolve are:
> 
> * Where do we take the contracts from dxVote. In the config we have all networks tokens and contracts, but I think we look contracts in dxVote according to this issue specifications. which are those?
> * styling issues (showing some screenshots) (I don't want to take my own decision here, since seems we look for consensus)
> 
> This is a dxDao token found on the config for all this addresses (a test in where all the addresses are a token address) Screen Shot 2021-10-08 at 11 51 30 AM
> 
> This is an AAVE token found on the config for all this addresses (a test in where all the addresses are a token address) Screen Shot 2021-10-08 at 12 41 34 PM
> 
> This are addresses resolved to an ENS (a test in where each address resolves to smolpp.eth) Screen Shot 2021-10-08 at 11 51 45 AM

Why are we seeing token icons ther everywhere? Looks like tokens are voting, but I looked at the code and it looks good, can you remove the token icon display for now?

If it is an ERC20 token and it is registered show TokenSymbol (ERC20), it would be like `DXD (ERC20)`

rossneilson:
> Why are we seeing token icons ther everywhere? Looks like tokens are voting, but I looked at the code and it looks good, can you remove the token icon display for now?

So if I'm not mistaken I think this is just a visualisation of what this would look like for testing but in reality it would only ever appear when transferring or approving that token?
I think having "2000 DXD (LOGO) transferred to augusto.eth" would make sense, no?

rossneilson:
Found a few places missing translations to ens names. This proposal on arbitrum is a great example where there are a few ens names being used. 
On the right side it all looks correct, but on calls and history it is not being translated, can we add this here?

0xa5A29f81EEE450eC189b2F8B4562af1785595D69 = 0xklom.eth
0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B = rossdev.eth

https://deploy-preview-199--dxvote-dev.netlify.app/?utm_source=github&utm_campaign=bot_dp#/arbitrum/proposal/0x6a35735ac54fead6f01784165381d3861f1eeae48846f949c6bf4af3749c5943

スクリーンショット 2021-10-14 12 34 01

luchux:
@rossneilson cool! will check into those now.

luchux:
an update of whats shuldd be seen by reviewers.

screencapture-localhost-3000-2021-10-14-06_03_36

src/pages/Proposal.tsx
```diff
@@ -143,6 +141,58 @@ const ProposalCallText = styled.span`
   white-space: pre-line;
 `;
 
+const Vote = styled.div`
+  display: flex;
+  font-size: ${({ theme }) => theme.votes.fontSize};
+  justify-content: space-between;
+
+  > * {
```
rossneilson, line 149: What is this?

src/pages/Proposal.tsx
```diff
@@ -824,13 +873,15 @@ const ProposalPage = observer(() => {
...
         votingMachineTokenApproved.toString() === '0' ? (
           <SidebarRow>
-            <small>Approve {votingMachineTokenName} to stake</small>
-            <ActionButton
-              color="blue"
-              onClick={() => approveVotingMachineToken()}
-            >
-              Approve {votingMachineTokenName}
-            </ActionButton>
+            <div style={{ display: 'flex', flexDirection: 'column' }}>
```
rossneilson: I think wrappers and styled components etc are better than inline styles even for small things like this

src/utils/address.ts
```diff
@@ -21,8 +38,64 @@ export function toChecksum(address) {
...
+  }
+  return tokenObject;
 }
+
+export function getDxVoteContract(address) {
+  let contract;
+  let networks = Object.keys(appConfig);
+  for (let i = 0; (i < networks.length && !contract); i++) {
+    let contracts = appConfig[networks[i]]?.contracts;
+    for (let [key, value] of Object.entries(contracts)) {
```
rossneilson: Any reason why we dont use es6 arrays functions here?
Something like: `Object.entries(contracts).forEach(() => {})`

src/pages/Proposal.tsx
luchux: every children

src/pages/Proposal.tsx
luchux: yes yes! good catch

src/utils/address.ts
luchux: oh yeah, mental note of redo later never worked. Good Catch! fixed!

src/pages/Proposal.tsx
luchux: I would love to change a lot of the markup in Proposal to make it more maintainable. 
Probably with History and Calls modification will introduce a nicer  manage of the Boxes and its styles.

luchux:
@hhamud seems your changes are marked as unresolved, but i did change them and marked as resolved. worth re/reviewing.

hhamud (approved):
LGTM!

PR #202, opened 2021-09-30
Hotfix/198 undefined ifps service

luchux:
done: 
- wrap the ifps.get method inside a try/catch.


hhamud (approved):
I was having the same issue,  good hotfix.

AugustoL (approved):

PR #206, opened 2021-10-01
Convert call data to multicall inputs

### LINKED ISSUE #182: Convert call data to multicall inputs
**Problem Description**
When displaying a multicall proposal, DXvote now shows the call data and the contract it's interacting with. On Alchemy, it shows you all of the inputs so that you can verify the proposal. F

or the buyback proposals, that would have the order info. On alchemy it's raw data (0x and lots of decimals) but it would be nice to have DXvote display that and token names/regular amounts, so that voters can verify. 

DXvote:

Screen Shot 2021-09-27 at 3 01 38 PM

vs. Alchemy
Screen Shot 2021-09-27 at 3 05 57 PM

hhamud:
* Closes #182 
Done:
* created `useEtherscanService` and `useABIService` hooks
* removed HTML from `daoService`
* refactored and moved `decodedWalletScheme` to `useABIService` hook

Todo:
* Expand ABI call to all DXDao bases, (arb testnet, xdai, etc)
* UI of call data
* Error handling

Question:
imagewould organising the params info in a table be best?. e.g. column with parameter name, type and then value.


src/pages/Proposal.tsx
```diff
@@ -262,10 +263,12 @@ const ProposalPage = observer(() => {
...
       : bnum(0);
-
-  let proposalCallTexts = new Array(proposal.to.length);
+  
+const proposalCall = async () => {
+  const proposalCallArray = []
   for (var p = 0; p < proposal.to.length; p++) {
-    proposalCallTexts[p] = daoService.decodeWalletSchemeCall(
+    
+    proposalCallArray[p] =   await daoService.decodeWalletSchemeCall(
```
luchux: 2 spaces before await

hhamud: I'll remember to format the document once its finished

src/components/CallDataInformation.tsx
```diff
@@ -0,0 +1,196 @@
...
+import { ProposalCalls, useABIService } from 'hooks/useABIService';
+import { observer } from 'mobx-react';
+import { useEtherscanService } from 'hooks/useEtherscanService';
+
+const CallDataInformation = observer(({ advancedCalls }) => {
+  const {
+    context: { daoStore, configStore },
+  } = useContext();
+
+  const proposalId = useLocation().pathname.split('/')[3];
```
rossneilson: To make the component more generic this could be a prop and the logic to find the id from URL should be moved to the parent.

src/components/CallDataInformation.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const scheme = daoStore.getScheme(proposal.scheme);
+  const { decodedCallData, ABI } = useABIService();
+  const { getContractABI, loading, error } = useEtherscanService();
+  const [ProposalCallTexts, setProposalCallTexts] = useState<ProposalCalls[]>(
+    new Array(proposal.to.length)
+  );
+
+  const proposalCallArray = [];
+  const getProposalCalls = async () => {
+    for (var p = 0; p < proposal.to.length; p++) {
```
rossneilson: Maybe this should be a `proposals.forEach((proposal, index) => {})` loop for better readability?
Or at least renaming `p` to something like `index` may be better

src/components/CallDataInformation.tsx
```diff
@@ -0,0 +1,196 @@
...
+      );
+      setProposalCallTexts(proposalCallArray);
+    }
+  };
+  useEffect(() => {
+    getProposalCalls();
+  }, []);
+
+  if (loading) {
+    return <div>Loading...</div>;
```
rossneilson: We should use spinners throughout the UI, there is one which should work in components/common/PendingCircle.tsx

src/components/CallDataInformation.tsx
```diff
@@ -0,0 +1,196 @@
...
+            </div>
+          );
+        })}
+      </div>
+    );
+  }
+
+  return (
+    <div>
+      {ProposalCallTexts.map(
```
rossneilson: A general note and please let me know if you disagree, but I think it is better to construct this jsx outside of the main return function.
So for example moving them into separate functions which construct components into an array which is then rendered here instead. 
This makes the main render a lot more readable as well as making for easier splitting of components in future if the need arises.

src/hooks/useABIService.ts
```diff
@@ -0,0 +1,196 @@
...
+                normalizeBalance(
+                  callParameters[paramIndex],
+                  recommendedCallUsed.params[paramIndex].decimals
+                ) +
+                '</italic>'
+            );
+          else
+            decodedCallText = decodedCallText.replaceAll(
+              '[PARAM_' + paramIndex + ']',
+              '<italic>' + callParameters[paramIndex] + '</italic>'
```
rossneilson: Not major but using `TextString${variable}moreText` may be easier here

rossneilson (commented):
A few comments from reading the code, the preview seemed to fail however.

Running it locally seems to work on mainnet however it looks like all other networks have issues. 
Integrating with all explorers for rinkeby, xdai and arbitrum may be difficult for this PR but at the very least it should default back to how it displayed the data before as currently to users of other networks this will look like a downgrade.

rossneilson:
We should also consider having a small error message for when the etherscan call fails. Perhaps one that could prompt user's to enter an etherscan API key. 
Also we should, maybe have a default key, not sure on how quickly we would hit limits though.

rossneilson:
It seems the missing API key error is hidden by default until the advanced view option is selected. And so it appears no calls are being made at first

Seems to be an issue in general with calls not showing up without pressing advanced icon

A solution should be for the basic info to be in basic and things like bytecode should be in advanced

スクリーンショット 2021-10-14 10 06 57

rossneilson:
Something that could be done in another PR if need be but I think is high priority is adding some redundancy here.
If we have reliance on a centralised explorer for fetching ABIs I think we need a back up.
Could you add some local ABIs in json files for the most used contracts etc so we dont need to rely on fetching them for good UX.

rossneilson:
Also I guess we need a way of setting different block explorer APIs in the settings config.
Currently it just shows etherscan and I think it needs overwitten each time right? We should store the network api keys separately and ideally have a link to where they can be found in each block explorer's API.

Essentially we need to think about any user coming in and wondering why they cant see proposal details, currently I think they path to most people getting it working is very difficult.

hhamud:
> Something that could be done in another PR if need be but I think is high priority is adding some redundancy here. If we have reliance on a centralised explorer for fetching ABIs I think we need a back up. Could you add some local ABIs in json files for the most used contracts etc so we dont need to rely on fetching them for good UX.

the flow is augusto's recommended calls --> blockexplorer ABI fetch --> display raw data. The most used ABI's should already be in dxvote and the etherscan calls are just there for any calls to contracts that fall outside of them.

hhamud:
> Also I guess we need a way of setting different block explorer APIs in the settings config. Currently it just shows etherscan and I think it needs overwitten each time right? We should store the network api keys separately and ideally have a link to where they can be found in each block explorer's API.
> 
> Essentially we need to think about any user coming in and wondering why they cant see proposal details, currently I think they path to most people getting it working is very difficult.




Yeah this should be a separate issue.

hhamud:
> It seems the missing API key error is hidden by default until the advanced view option is selected. And so it appears no calls are being made at first
> 
> Seems to be an issue in general with calls not showing up without pressing advanced icon
> 
> A solution should be for the basic info to be in basic and things like bytecode should be in advanced
> 
> スクリーンショット 2021-10-14 10 06 57

I'll get to that

rossneilson (commented):
Could we add back in the call number annotation (Call #1, Call #2) as well as a break between the calls.
Currently looking at a multicall its a lot better seeing functions and proper params, but with more than 1 call they currently merge together.
I also think bolding the param names could make things clearer.
Now:
スクリーンショット 2021-10-16 13 04 10
Before:
スクリーンショット 2021-10-16 13 04 20

rossneilson:
Also a spinner would be great as currently with 1+ calls it loads them one by one and if the second block explorer API call takes a while it looks like there is only 1 call.

Ideally the spinner should display loading 1/2 calls, then loading 2/2 calls.

Also loading all calls at the same time may make sense too for better performance (Promise.all).

rossneilson (approved):

PR #214, opened 2021-10-04
hotfix/advancedCallTo

### LINKED ISSUE #201: In advanced call, cannot change "to" value of first call
**Description**
When trying to create a proposal to send test SWPR on RInkeby to a multisig, I found a bug. I wanted to use advanced calls, and in particular a single call to the test SWPR token with an encoded `transfer` as calldata. Issue is that the first call in the list always had the `to` value set to `0xb05d148c6A9d9C0eb3fE0A091a68d6DeDac63f3b` (which should be the controller). I've tried a couple of things:

- Changing the value manually in the input's element is not working since when clicking "upload to IPFS" the value gets set back to the original one
- Trying to add a new call with the correct `to` value and then delete the old one does not work. Doing so SEEMS to work initially because it lets you set the correct `to` address, but when trying to delete the old one by clicking on the x sign, the new one is deleted instead.

hhamud:
* closes #201 
* removed change to call address that points to the controller address
* Note: not too sure the purpose of this if statement that I just removed.

### DIFF
--- a/src/pages/NewProposal.tsx
+++ b/src/pages/NewProposal.tsx
@@ -564,8 +564,7 @@ const NewProposalPage = observer(() => {
       setCallsInState(calls);
     }
   }
-  if (calls[0] && calls[0].allowedFunctions.length === 0 && allowedToCall[0])
-    onToSelectChange(0, allowedToCall[0].value);
+
 
   return (
     <NewProposalFormWrapper>


rossneilson:
I'll take a look into what this if statement does this afternoon but we shouldnt be removing code just because it fixes something, we need to properly understand what code does before changing it. If we dont do this we will find ourselves only fixing bugs and not building any new features.
As a side note, this is somewhere that comments would be great.
Maybe @AugustoL can chime in on what the if statement is for?

hhamud:
> I'll take a look into what this if statement does this afternoon but we shouldnt be removing code just because it fixes something, we need to properly understand what code does before changing it. If we dont do this we will find ourselves only fixing bugs and not building any new features. As a side note, this is somewhere that comments would be great. Maybe @AugustoL can chime in on what the if statement is for?

Yeah I know its a cardinal sin but I wasn't able to figure out why we allow advanced calls to any address but then overwrite the `to` field for the first call which we should never do without alerting the user first.

AugustoL (approved):

PR #216, opened 2021-10-05
fix: Removing additional min stake in proposals page

### LINKED ISSUE #210: Double counting of GEN/staking token in proposal view
**Description**
On the proposal page, there is a double counting of the GEN/staking token. If you look on individual proposals, it's the right amount. I think it's actually only the downstake. It should be 200 GEN, but it shows up as 400 GEN

Screen Shot 2021-10-01 at 5 32 44 PM
Screen Shot 2021-10-01 at 5 33 36 PM

rossneilson:
It seems the min downstake is already included in the downstakle amount and so by adding it back we ad this bug of it showing double the amount on the proposals page.

Let me know if I missed something and there is a case where we need this but in testing I could not find any.

Closes #210

### DIFF
--- a/src/pages/Proposals.tsx
+++ b/src/pages/Proposals.tsx
@@ -335,9 +335,6 @@ const ProposalsPage = observer(() => {
                 (schemeFilter === 'All Schemes' ||
                   proposal.scheme === schemeFilter)
               ) {
-                // const minimumDaoBounty = daoStore.getVotingParametersOfProposal(
-                //   proposal.id
-                // ).minimumDaoBounty;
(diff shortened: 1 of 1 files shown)


AugustoL:
Now it shows the right downstake value in all networks?

rossneilson:
Yes they now match proposals and proposal page and are the same as on alchemy also

luchux (approved):
@rossneilson seems legit to me.

luchux (approved):
legit!

AugustoL (approved):

PR #217, opened 2021-10-05
Feature/network switcher

### LINKED ISSUE #89: Make proposals visible without connecting to Metamask
**Is your feature request related to a problem? Please describe.**
I usually check daotalk posts from my phone and when someone links their onchain proposal, I am unable to open them to check.

**Describe the solution you'd like**
We need to have a disconnected state where users can view proposal details even when they dont have metamask installed. 
Users can browse through proposals in the disconnected state and if needed, can connect to Metamask to upvote / downvote / boost etc..

image

### LINKED ISSUE #154: Implement Network Switcher
We currently do not have a way to switch network from inside DXvote. While not so important with MetaMask, WalletConnect needs this for switching chains.

mprasanjith:
- This adds a network switcher UI and the logic for switching the network.
- If an injected wallet user hasn't connected to the network before, a request is sent for the user to add it.
- How routing is handled has been refactored to make sure network switching works when no wallet is connected.
- This code change also includes a refactor to how we keep the network configurations. I've brought everything to a single location, for easier maintenance in the future.
- The default RPC has been changed to Alchemy.

image


src/components/NetworkModal/index.tsx
```diff
@@ -0,0 +1,144 @@
+import { useEffect, useState } from 'react';
+import styled from 'styled-components';
+import { observer } from 'mobx-react';
+
+import { Modal } from '../Modal';
+import Option from './Option';
+import { getChains } from 'provider/connectors';
+import { useContext } from '../../contexts';
+import { useRpcUrls } from 'provider/providerHooks';
+import { InjectedConnector } from '@web3-react/injected-connector';
```
luchux: Beautiful work! 

should we keep imports order as:

Libs (react, mobx, etc)
Components 
Utils 
Styles

?
seems a nice practice consistent along many es6 projects.  is a sugar touch, just curious...

src/components/Web3ConnectStatus/index.tsx
```diff
@@ -51,52 +52,62 @@ const Web3ConnectStatus = observer(props => {
...
+    } else if (chainId) {
       return (
         <div style={{ display: 'flex' }}>
-          <AccountButton onClick={toggleWalletModal}>
-            {shortenAddress(account)}
-          </AccountButton>
           <AccountButton
-            onClick={toggleWalletModal}
+            onClick={toggleNetworkModal}
             style={{ fontSize: '14px' }}
```
luchux: may be you would like to extend this AccountButton and remove this inline style?

src/utils/constants.ts
```diff
@@ -9,23 +13,122 @@ export const ERC20_TRANSFER_SIGNATURE = '0xa9059cbb';
...
-  '1': `https://mainnet.infura.io/v3/${defaultInfuraKey}`,
-  '4': `https://rinkeby.infura.io/v3/${defaultInfuraKey}`,
-  '100': 'https://rpc.xdaichain.com/',
-  '42161': `https://arbitrum-mainnet.infura.io/v3/${defaultInfuraKey}`,
-  '421611': `https://arbitrum-rinkeby.infura.io/v3/${defaultInfuraKey}`,
-};
+
+export const NETWORKS: {
+  id: number,
+  name: string,
```
luchux: wouldn't this be better defined as a type or an interface, and then 
export const NETWORKS: networkType [] ?

src/utils/constants.ts
```diff
@@ -9,23 +13,122 @@ export const ERC20_TRANSFER_SIGNATURE = '0xa9059cbb';
...
+    defaultRpc: `http://localhost:8545`,
+    nativeAsset: {
+      name: "Ethereum",
+      symbol: 'ETH',
+      decimals: 18
+    },
+  },
+];
+
+export const NETWORK_NAMES: { [chainId: string]: string } = NETWORKS.reduce(
```
luchux: and each of this functions get a subtype of the types deifned before?

luchux (commented):
Excellent work! just let some small comments...

src/components/NetworkModal/index.tsx
mprasanjith: Thanks for spotting this. We definitely should have a proper ordering. I'll update this PR.

Is there any way to have a linting rule for this?

src/utils/constants.ts
mprasanjith: That's very elegant!

mprasanjith: Just to make sure we're talking about the same thing, in this scenario the typing will look like:
```
const NETWORK_NAMES: Record<ChainConfig['id'], ChainConfig['name']> = NETWORKS.reduce(
```
Or is there any other way of typing this?

luchux: yeah thats a way! another way I have seen is: 

```
type ChainID { 
   id: string
}
   
type Chain  = ChainID & {
  ....
}
```

useful to use then ChanID for the repeated params of several functions.

AugustoL (approved):

rossneilson (approved):

PR #218, opened 2021-10-06
realigning proposals page with grid and old layout

luchux:
This is a proposal PR to bring back the old layout so we can add an extra column in the table without loosing space. 
Also this layout offers a better alignment of the ProposalFilter with respect to the table and the visual semantic relation of the search and filter actions with the data that is searched/filtered. 

The layout is simplified with a vertical grid 20% 80%
and the filters simplified a flex row with last item (NewProposalButton with self align to the right). 
Screen Shot 2021-10-06 at 1 42 33 AM

### DIFF
--- a/src/PageRouter.tsx
+++ b/src/PageRouter.tsx
@@ -3,7 +3,7 @@ import { observer } from 'mobx-react';
 import { useContext } from './contexts';
 import { FiZapOff, FiZap } from 'react-icons/fi';
 import { useLocation, useHistory } from 'react-router-dom';
-import Box from './components/common/Box';
+import { Box } from './components/common';
 
 const PageRouterWrapper = styled.div`
   margin-top: 20px;
--- a/src/components/AccountDetails/index.tsx
+++ b/src/components/AccountDetails/index.tsx
@@ -56,19 +56,10 @@ const InfoCard = styled.div`
 
 const AccountGroupingRow = styled.div`
   ${({ theme }) => theme.flexRowNoWrap};
-  justify-content: space-between;
+  justify-content: space-around;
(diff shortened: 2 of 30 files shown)


mprasanjith:
On the Netlify preview, I'm seeing this.

image

Is the extra whitespace on the left and right intentional?

luchux:
> On the Netlify preview, I'm seeing this.
> 
> image
> 
> Is the extra whitespace on the left and right intentional?

weird mate, im seeing this in the preview:

luchux:
> On the Netlify preview, I'm seeing this.
> 
> image
> 
> Is the extra whitespace on the left and right intentional?

weird @mprasanjith  this is my netlify preview result: 

Screen Shot 2021-10-06 at 3 08 58 AM

mprasanjith:
> weird @mprasanjith this is my netlify preview result:
> 
> Screen Shot 2021-10-06 at 3 08 58 AM

Weird indeed. Checked on both Firefox and Edge, and getting the same issues on a 16:9 display.

rossneilson:
Hey Lulli, so we have some plans for the UI to make more use of the left hand sidebar space and so I dont think this is a great design decision. We should instead be trying to show only the data we want in a more efficient way. Also for the planned card view this layout does not work very well.
Also I guess this is more personal preference but I find our current design to look far more modern.

We also have a huge number of other areas where we need to improve UI and UX and so we should focus more on those areas instead of constant iteration here.

luchux:
@rossneilson sure thing! 
I will discuss options of the timestamp column with the current layout then. Sorry, and thanks for the feedback.

PR #220, opened 2021-10-06
Updates scoring weights

fluiddrop:
Changes a vote on the passing state, to attribute 2 points, while increasing the minimum amount of points for a proposal creation to 3. Logically, passing a successful proposal should award more points than simply voting.


AugustoL (commented):
The issue with increasing the minimum score points earned by creating a proposal is that there is no limit to the amount of proposals that can be created. It is going to be expensive to execute this "attack", but it can happens anyway.

Also the thing we cared the most and what we want to reward the most are the votes set on the final decision. We shouldn't be decreasing this number IMO.

PR #222, opened 2021-10-07
feat: Added calls to new swapr campaign creation template

### LINKED ISSUE #219: Automatic filling of SWPR farming campaigns in proposal creation
Currently we have the expert mode UI for all proposals. Long term this will be replaced by our new UX flow for various proposal types however as we have a security need for more simplicity doing these proposals frequently and right now this needs a short term fix.

Ideally we want a proposal template that can select default calls in the create proposal screen. Currently these can only set description templates but we should be able to auto populate some calls also.

For swapr farming campaigns this would be the approve call for swpr token and liquidity mining campaign call. 
Any other automation here would be great for security also.

(Longer term solution with new UX flow will also be high priority after this issue is solved to add even more security around these proposals that we often have human error issues with currently)

rossneilson:
Requested by @jpkcambridge we now have automatic filling of calls when selecting a template.
This is optional and only currently implemented for swapr campaign creation in an attempt to reduce human error and make new proposal creation for fortnightly epochs simpler.

Closes #219

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -240,6 +240,25 @@
         "name": "Signal Proposal",
         "title": "Signal Proposal for [DAO_ACTION]",
         "description": "## Description \n\n## Proposition \n\n## Action Plan"
+      },
+      {
+        "name": "SWPR Farming Campaign",
+        "title": "[PAIR X/Y] Farming Campaign [DATE - DATE]]",
+        "description": "Simple description",
+        "calls": [
+          {
+            "functionName": "createDistribution(address[],address,uint256[],uint64,uint64,bool,uint256)",
+            "params": [
+              {"id": "rewardsTokenAddresses", "defaultValue": "0xecA7F78d59D16812948849663b26FE10E320f80C"},
+              null,
+              null,
+              null,
+              null,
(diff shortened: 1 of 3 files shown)


AugustoL (approved):
Looks alright, but John should test it since he requested it. summoning @jpkcambridge here

src/config.json
```diff
@@ -240,6 +240,31 @@
...
+          {
+            "to": "0xdE903E2712288A1dA82942DDdF2c20529565aC30",
+            "functionName": "approve(address,uint256)",
+            "params": ["0xecA7F78d59D16812948849663b26FE10E320f80C",""]
+          },
+          {
+            "to": "0xecA7F78d59D16812948849663b26FE10E320f80C",
+            "functionName": "createDistribution(address[],address,uint256[],uint64,uint64,bool,uint256)",
+            "params": [
+              "0xecA7F78d59D16812948849663b26FE10E320f80C",
```
jpkcambridge, line 258: I think this should actually prepopulate as an array, and instead of the staking rewards campaign factory address, it should be the Arbitrum DXD address, which is `0xdE903E2712288A1dA82942DDdF2c20529565aC30`

Issue #227, opened 2021-10-07
UI issue in proposals list when filtering


mprasanjith:
When filtering proposals on the list view, if there aren't many proposals to show the rows get spread out, and looks like this:

![image](https://user-images.githubusercontent.com/25136207/136429408-d508134e-61c2-4afb-9a4f-1fdac480ec68.png)

If there are enough proposals to overflow the default height of the container, this issue does not happen.

luchux:
Im gonna fix this

luchux:
Fixed here: https://github.com/DXgovernance/dxvote/pull/229

PR #229, opened 2021-10-08
removing duplicate footer when in proposals, and uplifiting table whe…

luchux:
Fixed: 
-  a duplicate Footer when in /proposals (in the left column, and at the end of the page). Did a condition rendering of footer based on the route. 
- Issue https://github.com/DXgovernance/dxvote/issues/227 when few rows in table, rows were taking big space to complete the page.


src/index.tsx
```diff
@@ -40,6 +40,51 @@ function getLibrary(provider) {
...
+        {' '}
+        <ForumPage />{' '}
+      </Route>
+      <Route exact path="/faq">
+        {' '}
+        <FAQPage />{' '}
+      </Route>
+      <Route exact path="/:network/proposals">
+        {' '}
+        <ProposalsPage />{' '}
```
AugustoL: Cant we just have the `<footer/>` element under the proposals page element instead of checking the location url?

luchux: @AugustoL the problem is that <Footer> has been added inside the Proposals page because of the layout that was created (at the bottom of the left sidebar in /proposals). 
In such case, /proposals renders 2 footers, one from the router and one from the Proposals page. 

only way to only show the sidebar one (the one render in Proposals page) is to disable the one rendered in the router.

AugustoL (approved):

rossneilson (approved):

rossneilson:
Fixed some merge conflicts will merge shortly as I want the footer stuff :P

PR #232, opened 2021-10-08
Feature/proposal types

### LINKED ISSUE #187: Choose proposal type view
For a better UX we should have a clear page where the user selects the kind of proposal they want to make, allowing us to steer them to the correct flow without confusing the user with the somewhat intimidating current UI we have for every type of proposal currently.

This page should be accessible upon pressing "Create proposal" from the main screen

Initially we will have 2 options, one for 'contributor proposal' and one for 'custom' with plans to add more soon.
In future we may decide that the boxes should be smaller when we have lots of options so this should be easily scalable.
Currently I think just the title will do fine but the options could be colour coordinated to their appropriate wallet schemes also.

**These options MUST be configurable from a simple config file**

Upon clicking custom the user should be taken to the existing proposal creation view and any other options should not do anything right now.

The screenshots and figma are meant more as wireframes than exact design specifications so some choice lies with the developer on how they should look, feel free to ask for feedback from everyone.
The boxes could also have hover animations scaling and changing the drop shadow as they are moused over

スクリーンショット 2021-09-28 9 27 26

https://www.figma.com/file/N4e2YuSiv1SycIhmqhIXGB/DXgov?node-id=95%3A2

### LINKED ISSUE #224: Split up configs into networks
Currently we have a very large 2.5k line config file and it would be better if this was split up into networks at the very least.

In future we can maybe further split up these configs but for now this should be a good first step.

src/configs/[network]/config.json

rossneilson:
A basic new page for the new UX flows.
Currently nothing directly links here but once the entire flow is complete we can change the new button on the main page.
Swapr and contributor proposal link to broken pages but custom goes to the current new proposal page which will soon become the power user only UI. 

Hovering over has some basic animations.
Any additional design feedback would be great but can also be iterated on in future. 

スクリーンショット 2021-10-08 11 13 11

Also split up the configs into network folders, this can be further split up in future if needed. The UI just creates one big file again since I couldn't find a simple way of dynamically importing the correct config. 

Closes #224 
Closes #187


src/index.tsx
```diff
@@ -69,10 +70,14 @@ const Root = (
                 {' '}
                 <ProposalsPage />{' '}
               </Route>
-              <Route exact path="/:network/new">
+              <Route exact path="/:network/create/custom">
```
AugustoL: maybe this route can take a wildcard parameter in /create? so we dont have have a link for each proposal type? `/create/:proposalType` could work.

rossneilson: Done, changed urls around a bit

We will have:
create/type
create/metadata/:proposalType
create/submit/:proposalType

AugustoL (approved):

PR #239, opened 2021-10-11
Feature/metadata page

### LINKED ISSUE #188: New proposal metadata page
**Blocked by #187**

In the simplified UX flow for proposal creation we will have a first view after selecting proposal type be a metadata editor.
For the scope of this issue this includes the ability to edit the title and description via a markdown editor with more features to be added later.

This first implementation should be accessible upon clicking the "Contributor proposal" option. (Not CUSTOM)
This page should be paramterised and reusable, multiple different types of proposal will use this to edit metadata before going on to different next pages and so at the very least proposalType must be passed through to know where to navigate the user afterwards. (For now the page should not lead anywhere)

The screenshots and figma are meant more as wireframes than exact design specifications so some choice lies with the developer on how they should look, feel free to ask for feedback from everyone.
Basic design should follow the screenshots having a lefthand sidebar for set inputs and a more freeform MD editor with preview on the right.

スクリーンショット 2021-09-28 9 38 04

https://www.figma.com/file/N4e2YuSiv1SycIhmqhIXGB/DXgov?node-id=95%3A2

rossneilson:
Added new metadata page to be used for new proposal creation UX flow.
Simple for now but more will be added to sidebar in future.
Not used until next issue implements the final submission page for a proposal type.

Height of markdown editor is a bit janky, at some point we should maybe be upgraded to a better md editor package.

Close #188

スクリーンショット 2021-10-11 16 25 05


src/pages/Metadata.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+export const CreateMetadataPage = observer(() => {
+  const history = useHistory();
+  const { proposalType } = useParams();
+
+  const [descriptionText, setDescriptionText] = useState(
+    localStorage.getItem('dxvote-newProposal-description') || ''
+  );
+  const [title, setTitle] = useState(
+    localStorage.getItem('dxvote-newProposal-title') || ''
```
mprasanjith, line 92: I think we need to persist the proposal title to the local storage.

src/pages/Metadata.tsx
```diff
@@ -0,0 +1,120 @@
...
+  border: 1px solid #10161a33;
+  border-radius: 3px;
+  padding: 20px 10px;
+  height: 80vh;
+  overflow: scroll;
+`;
+
+export const CreateMetadataPage = observer(() => {
+  const history = useHistory();
+  const { proposalType } = useParams();
```
mprasanjith, line 79: Curious if we should introduce a validation to the proposalType. Any random string in the URL works right now.

src/pages/Metadata.tsx
rossneilson: Missed this, thanks!

src/pages/Metadata.tsx
rossneilson: Oh yeah, will do this, thanks for suggestion

mprasanjith (approved):

PR #240, opened 2021-10-11
Filtering Proposals by Status and Scheme

luchux:
Filter and sort proposals initially with the following requirements: 

Proposals in Queued or QuietEndingPeriod status with more than 10% of positive votes. (Ordered from time to finish, from lower to higher)
Proposals Boosted. (Ordered from time to finish, from lower to higher)
Proposals Pre Boosted.
Proposals in Queued or QuietEndingPeriod status with lower than 10% of positive votes. (Ordered from time to finish, from lower to higher)
Proposals in Executed status. (Ordered in time passed since finish, from lower to higher)

If Status filtering applied, filter by that status. 
if Scheme filtering applied, also filter by schema. 
if Title search, also filter by title. 
[in a chained filtering fashion]

Screen Shot 2021-10-12 at 6 06 40 PM
Screen Shot 2021-10-12 at 6 06 32 PM
Screen Shot 2021-10-12 at 6 06 22 PM
Screen Shot 2021-10-12 at 6 06 09 PM


luchux:
@rossneilson @AugustoL 
So some comments on this solution: 
- I am applying the filterInitialCriteria function for filtering and sorting and described by AugustoL in this feature. 
- I am generating the State Filter <option>'s from the enum VotingMachineProposalState now. Should we provide more than those? 
- Schema filter, and search by title is applied after the result of State filter:  If 'Any Status', those schemes and title mach from the initial criteria filtering. If Status is selected, scheme and title filter from those results. 
- Can we move the filterInitialCriteria function to a helper? I am using the repAt to calculate the % of votes, so we need access to the daoStore from the helper if we do so.  (or can we calculate that % posiitves,  from the data in the proposal object ? without the repAt? ) 
- I removed any type of filter at Component rendering Level, (it was a filter when iterating over the proposals and rendering each row). Now the whole array gets rendered, and the that array is the one that is filtered in each criteria (no more conditional rendering). 
- I see another PR that touched a lot of the title search filtering https://github.com/DXgovernance/dxvote/pull/244/files, I can manage to merge this filtering title way with my changes.

src/pages/Proposals.tsx
```diff
@@ -157,40 +159,82 @@ const ProposalsPage = observer(() => {
...
-   * Preboosted are ordered in boostTime and not in Finish Time.
-   *
-   */
-  const sortedProposals = mapEnum(VotingMachineProposalState, p => {
-    return (
-      allProposals
+  //filtering and sorting proposals for All States criteria
+  const filterInitialCriteria = (proposals) => {
+    // (QuitedEndingPeriod || Queded) && positiveVotes >= 10% (Ordered from time to finish, from lower to higher)
+    let earliestAbove10 = proposals.filter(
```
AugustoL, line 181: Can we take this value form a constant in utils? Somehing like `QueuedPriorityThresold`, so we can change it later easily in case we want to.

src/pages/Proposals.tsx
```diff
@@ -157,40 +159,82 @@ const ProposalsPage = observer(() => {
...
-  if (sortedProposals.length > proposals.length) setProposals(sortedProposals);
+    boosted.sort(orderByNewestTimeToFinish);
+
+    let preBoosted = proposals.filter((proposal: Proposal): Boolean => proposal.stateInVotingMachine === VotingMachineProposalState.PreBoosted);
+    preBoosted.sort(orderByNewestTimeToFinish);
+      
+    // (QuitedEndingPeriod || Queded) && positiveVotes < 10% (Ordered from time to finish, from lower to higher)
+    let earliestUnder10 = proposals.filter((proposal: Proposal): Boolean => {
+          
+      // is it necessary to all this datoSTore.getRepAt?
```
AugustoL: Yes! Because this is the total supply of the REP token when that proposal was created. and based on that and the REP used in votes we calculate the percentage of votes.

AugustoL (changes requested):

src/pages/Proposals.tsx
luchux: done! great call

src/pages/Proposals.tsx
luchux: OKOK!

src/utils/proposals.ts
```diff
@@ -341,3 +344,15 @@ export const decodeProposalStatus = function (
       };
   }
 };
+
+//a: Proposal, b: Proposal but boostTIme is not in Proposal interface. 
+//is it added in Cache?
```
AugustoL, line 349: Sorry I just saw this, yes it is, in Proposal.boostedPhaseTime. Do we need it though? Because we are calculating already the time that a proposal will finish based on their status and boostedPhaseTime before ordering, right?

src/utils/proposals.ts
```diff
@@ -341,3 +344,15 @@ export const decodeProposalStatus = function (
...
+
+//a: Proposal, b: Proposal but boostTIme is not in Proposal interface. 
+//is it added in Cache? 
+export const orderByNewestTimeToFinish = (a: any, b: any) =>
+  a.finishTime - b.finishTime;
+
+// finishTime is modified automatically to be TimeSinceFinished by 
+// the decodeProposalStatus function of utils/proposals.
+// so we may use the same function for everycase? 
+export const orderByNewestTimeSinceFinished = (a: any, b: any) =>
```
AugustoL: hmmm, not sure if I understand, but yes (? The finishTime received by the decodeProposalStatus gives us the time that the proposal is predicted to finish. And I guess that yes, we want to order the proposals based on their finish time sometimes.

Also this function does the same as the function above. And the variable is called `finishTime` so maybe lets just use only the function declared above `orderByNewestTimeToFinish` ?

AugustoL (commented):
I added a few comments, then looks we good.

src/utils/proposals.ts
luchux: @AugustoL yeah my bad here. Created 2 thinking for time passed since finished we had another param) 
"Proposals in Executed status. (Ordered in time passed since finish, from lower to higher)" ... the passed since finish is already calcuted in  finishTime of decodeProposalStatus for the executed ;) .

let 2 in case it was another field to look for. removing in a bit

luchux: good catch perrico ;)

src/utils/proposals.ts
luchux: oh for typescript params. Cause in some params we are accessing proposal.boostTime so we can't type that as Proposal, but may be a type for it that i didn't see. ;) will check it

luchux: if not will make the type extending Proposal and adding the missing fields

AugustoL (approved):

PR #254, opened 2021-10-14
Improve MetaMask UX

### LINKED ISSUE #242: Improve UX around MetaMask connectivity
When connecting MetaMask, the app switches to whatever network MetaMask is connected to and does a refresh. The user most probably want to interact with the page they're on, on the network the app is already connected to instead.

mprasanjith:
This PR changes the behavior of MetaMask and other injected wallets. Currently, the browser prioritizes the MetaMask connection, and eagerly connects to it whenever it's present. While this works generally, the issue is when someone wants to directly access a  proposal via a link, they have to first switch to the correct network on MetaMask manually.

This PR fixes this by only eager-connecting if the network is the same.
1. When visiting the DXvote homepage, by default it connects to mainnet.
2. The user is able to switch networks without switching it in MetaMask.
3. If MetaMask or other injected wallet is found, the active network of that wallet is checked.
4. If the wallet is on the same network DXvote is in, we will eagerly connect.
5. If not, the user is alerted to switch the wallet to the network DXvote is in.

image

**Sidenote**
- Web3 provider implementation in DXvote needs some refactoring and cleaning up. There's a lot of stuff we don't use, and there's some code duplication. Its better to revisit this later.

Fixes #242.

### DIFF
--- a/src/components/Web3ConnectStatus/index.tsx
+++ b/src/components/Web3ConnectStatus/index.tsx
@@ -2,10 +2,12 @@ import styled from 'styled-components';
 import { observer } from 'mobx-react';
 import { shortenAddress, toCamelCaseString } from '../../utils';
 import WalletModal from 'components/WalletModal';
-import { isChainIdSupported } from 'provider/connectors';
+import { getChains, injected, isChainIdSupported } from 'provider/connectors';
 import { useContext } from '../../contexts';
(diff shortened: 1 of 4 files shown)


rossneilson (approved):
This is fantastic Madusha!!! Will be a huge UX improvement and it works very smoothly

PR #261, opened 2021-10-14
Feature/209 optimization rep at

luchux:
each calculation of repAt is killer for the rendering performance. 

This PR is not meant to be in production, since this is a silly cache in the page state that keeps for i1Block -> repAt(i1Block). This is merely to demonstrate that the scalabilities issues of UI/UX for title filtering are not related to the library, but to the repAt calculation of each positive/negative.

What if we lightly cache in a state the blockAddrs and the repAt calculated during rendering in a light cache in state. 

Memoizing this function with useMemo doesn't work because we access it outside of functional components. 
THIS PR is based on https://github.com/DXgovernance/dxvote/pull/240 so only the last commit is actually this improvement.


luchux:
@mprasanjith the last commit here (this branch is based on an open PR that I have for fixing sort), is doing a cache of repAt in state. Memoizing this function didn't work with useMemo, and it has valid reasons to fail. We can discuss more, but this example shows that the problems are not in the indexing library neither in the render inefficiency of the page (fetching data, sorting, etc) but is a net issue with each call of repAt. 

I don't think this could be a solution per se, since we are caching value that may change in the time the user has loaded and is using the app, worth discussing with Augusto and Ross. 

but as you see, typing with this approach flies! 
Super keen to see if we can cache this at the daoStore level. my knowledge escapes a bit that.

luchux:
@mprasanjith 
I will close this PR. Talked to augusto in a better performance improvement. 
Will keep you updated soon.

PR #263, opened 2021-10-15
Make proposal.historyEvents to show blockchainLink in text.

luchux:
Given the way we construct the history, we can return the values injected in the string as params, and create a list of components of <span> <blockchainLink> for each part of the text.

### DIFF
--- a/src/pages/Proposal.tsx
+++ b/src/pages/Proposal.tsx
@@ -455,22 +455,22 @@ const ProposalPage = observer(() => {
     networkContracts.votingMachines.dxd &&
     networkContracts.votingMachines.dxd.address === votingMachineUsed;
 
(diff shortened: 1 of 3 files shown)


src/pages/Proposal.tsx
```diff
@@ -392,6 +392,20 @@ const ProposalPage = observer(() => {
     networkContracts.votingMachines.dxd &&
     networkContracts.votingMachines.dxd.address === votingMachineUsed;
 
+  const renderEvents = ({ text, text_params }) => {
+    let components_list = []
```
rossneilson: Any reason why this isnt camel case?

luchux: good catch

AugustoL (approved):

PR #267, opened 2021-10-15
Feature/table proposals header sticky

luchux:
Done: 
- Make the table header in proposals sticky, and make the whole table to be scrollable (so we don't hide headers, sidebar) when scrolling. 
- Fixed some coding format in other files for github actions check. 

Screen Shot 2021-10-15 at 12 24 05 PM


hhamud (approved):
LGTM!

PR #271, opened 2021-10-18
feat: Improve deployment behavior for IPFS.

### LINKED ISSUE #258: Fix IPFS deployments when opening PRs from forked repositories

mprasanjith:
This is a complete overhaul of the IPFS deployments GitHub actions we had previously.

With this change, the deployment actions are split into three different pipelines.
1. Build pipeline - Builds the frontend and archives the build output as an artifact.
2. Deploy pipeline - Deploys the latest build artifact into IPFS. Spawns automatically when the build pipelines finishes.
3. Cleanup pipeline - Cleans up the deployment when a PR is merged.

Previously we had an issue of the deployment failing when opening a PR from a fork. GitHub doesn't provide write access to the pipeline in this case for security reasons. The new setup works around this limitation by splitting the build and deployment into two pipelines. The build pipeline still runs with limited permissions, but spawns the deployment pipeline afterwards. This spawned pipeline has full privileges, and completes the deployment.

This PR also handles Environments in GitHub gracefully, and also leaves a comment when a deployment is completed successfully for good measure.

Caveats:
- Secrets are not provided to PRs from forks (intentionally). I suggest to move the Alchemy API key (which should be the only secret that is affected) to our environment configs in the code. It's not really a "secret" as its anyway exposed to the user.

Fixes #258.

**This PR needs to be merged into `master` for everything to work as expected. I'd hold off merging this until we're ready for the next release, as merging this will affect the current deployment setup.**

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -1,0 +1,43 @@
+name: Code Build
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [14.17.6]
+    steps:
+      - name: Inject workflow information
+        uses: potiuk/get-workflow-origin@v1_3
+        id: workflow-run-info
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Build dapp
+        run: CI=false yarn build
+
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ steps.workflow-run-info.outputs.targetCommitSha }}
+          path: build/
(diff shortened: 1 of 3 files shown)


rossneilson (approved):

.github/workflows/preview-build.yml
```diff
@@ -0,0 +1,43 @@
...
+        node-version: [14.17.6]
+    steps:
+      - name: Inject workflow information
+        uses: potiuk/get-workflow-origin@v1_3
+        id: workflow-run-info
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout
+        uses: actions/checkout@v2
```
AugustoL, line 26: We are missing the 
```
      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 0
          ref: ${{github.event.pull_request.head.ref}}
          repository: ${{github.event.pull_request.head.repo.full_name}}
          ```
Options to use head from the referenced pull request, not the one in source, right? I remember I had to add this to use the head from PR forks.

AugustoL: I think we should use it in all workflows that do checkout.

AugustoL (commented):
I added a comment.

Regarding alchemy.api key I can create one for production with a whitelist to be used only in dxvote.eth.link, we can add it as a env.production variable and run the build script using that.

Or maybe we can set the en.production variables in the source somewhere.

.github/workflows/preview-build.yml
mprasanjith: I'm listening to the `pull_request` event with limited privileges again, instead of `pull_request_target` so we won't need this.

AugustoL: I think this will fail then, since pull_request_target was changed in order to make it work from fork PRs. But merge and we can try,

mprasanjith: I have a workaround for that also, so this should work (hopefully 🤞). I got it working on my personal repo [here](https://github.com/mprasanjith/dxvote/pull/30) with these changes.

PR #273, opened 2021-10-19
Feature/225 git commit version

luchux:
Does: 
- calculates git commit in bash start.sh, build.sh, and dev.sh
- does show the commit and v in the footer (in prod only version) 
Screen Shot 2021-10-18 at 10 49 27 PM


rossneilson (approved):
Looks great!

PR #274, opened 2021-10-19
Allow user to enter apiKey for several blockexplorers

### LINKED ISSUE #266: Allow user to enter apiKey for several blockexplorers
Also I guess we need a way of setting different block explorer APIs in the settings config.
Currently it just shows etherscan and I think it needs overwitten each time right? We should store the network api keys separately and ideally have a link to where they can be found in each block explorer's API.

Essentially we need to think about any user coming in and wondering why they cant see proposal details, currently I think they path to most people getting it working is very difficult.

_Originally posted by @rossneilson in https://github.com/DXgovernance/dxvote/issues/206#issuecomment-943263516_

hhamud:
* closes #266 
* created `ethereumCheck` function that cuts out the api key portion of the api request for arbitrum and xdai chains. 

I have dug into the issue and it appears to be that the only block explorer sites that require an API key are the etherscan sites. Arbiscan and blockscout (xdai) does not need an API Key input.


src/utils/helpers.ts
```diff
@@ -75,3 +75,10 @@ export function parseCamelCase(text) {
   const parsed = text.replace(/([a-z])([A-Z])/g, '$1 $2');
   return parsed[0].toUpperCase() + parsed.substring(1);
 }
+
+export const etherumCheck = (networkName: string, etherscanAPIKey: string) => {
```
mprasanjith: Great work!

Maybe we could rename this function to something more clear? We can also use the [URLSearchParams API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString) for cases like this now, although I don't think its necessary to change it here.

hhamud: URL search params function would make it much easier to manipulate. I'll use that. thanks

src/services/EtherscanService.ts
```diff
@@ -1,40 +1,74 @@
...
+      module: 'contract',
+      action: 'getabi',
+      address: address,
+      apikey: appendEthAPIKey(networkName, this.etherscanAPIKey),
+    });
+    const getContractURL = new URL(
+      '/api?' + params.toString(),
+      NETWORK_APIS[networkName]
+    );
+    console.log(getContractURL);
```
mprasanjith: We can remove this console.log if not needed. Everything looks great!

mprasanjith (approved):
Saw a console..log we could remove

src/services/EtherscanService.ts
hhamud: Forgot to remove that. I'll do that before merge

PR #276, opened 2021-10-19
Updated loading screen

### LINKED ISSUE #255: Interesting loading screen
Our loading screen is a little bland right now and when it takes a while to load it can be misleading.

First we should remove the box and drop shadow since we no longer use it on proposals page.

We also need a spinner, this could be creative like a DXdao logo or something or just a normal grey one.

Next we should display exactly what is being done. I cant think of a good example but these are used often in video games to show what is being loaded or rendered during a loading screen. 
This could be "Getting cache from ipfs", "Parsing block 1314314", etc

Essentially something that is always updating to let the user know that nothing is broken when its actually just building the cache from on chain data.

mprasanjith:
- Adds a brand new loading screen when the UI is loading for the first time. It also handles failures in `BlockchainStore` when loading data gracefully.
- A new MobX store has been added for notifications. In the future, this can be reused when we want to implement toast notifications etc.
- We can reuse the error state of the loading screen when we want to alert the user to unrecoverable global errors. It might be beneficial to use this to setup a React error boundary on the root level of the app.
- While testing this, I also found a bug that prevented DXvote from loading when no injected wallet is present. This has also been fixed.

image

Fixes #255.

### DIFF
--- a/src/contexts/index.ts
+++ b/src/contexts/index.ts
@@ -19,6 +19,7 @@ import DaoStore from '../stores/DaoStore';
 import UserStore from '../stores/UserStore';
 import BlockchainStore from '../stores/BlockchainStore';
 import CustomRpcService from 'services/CustomRpcService';
+import NotificationStore from 'stores/NotificationStore';
 
 /*
 https://reactjs.org/docs/context.html#reactcreatecontext
@@ -40,6 +41,7 @@ export default class RootContext {
   configStore: ConfigStore;
   daoStore: DaoStore;
   userStore: UserStore;
+  notificationStore: NotificationStore;
   blockchainStore: BlockchainStore;
 
   abiService: ABIService;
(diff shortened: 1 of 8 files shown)


src/contexts/index.ts
```diff
@@ -19,6 +19,7 @@ import DaoStore from '../stores/DaoStore';
 import UserStore from '../stores/UserStore';
 import BlockchainStore from '../stores/BlockchainStore';
 import CustomRpcService from 'services/CustomRpcService';
+import NotificationStore from 'stores/NotificationStore';
```
luchux: Im just curious why BlockchainStore, and UserStore are imported from '../stores/' and NotificationStore from 'stores/' does it resolve the path well? 

And what about exporting all those in stores/index.ts  module so  here you can 
`import {UserStore, BlockchainStore, NotificationStore} from 'stores/'`

luchux (approved):
@mprasanjith excellent work. 
Tested locally and works like a charm, awesome new mobx states and sexy UI! 

just let a comment, that i don't see how is resolving well to load the store (cause the diff paths in the imports) but seems to be working well.

mprasanjith:
TypeScript now supports both types of imports (absolute from the project root and relative) out-of-the-box it seems. And new VS Code versions seem to default to a mix of using both. (The default is to use the shortest import.)

I went ahead and accepted what the IDE decided to import and didn't realize it was mixing the two up. Just pushed a change to fix it.

PR #280, opened 2021-10-19
Chore/update cache

AugustoL:
- This PR adds the new DXDVotingMachine deployed in xdai and permission registry into the xdai config and it will grab the newkly deployed quickwallet machine in the app when the app is in xdai.
- The cache was updated a few more blocks too.


AugustoL:
We need to merge this soon, xdai is broken in develop. hot fix needed. @rossneilson

rossneilson (approved):

PR #281, opened 2021-10-20
Fix/default provider setup

AugustoL:
- Adds default alchemy API key in env.production file. The RPC is whitelisted to only work with `dxvote.dev` and `dxvote.eth.link` domains.
- Fix bug that breaks the app when trying to load dxvote without any ethereum provider installed in the browser.


mprasanjith (approved):
I also spotted the injected wallets issue and did a [different fix](https://github.com/DXgovernance/dxvote/pull/276/files#diff-4b6b20d64fb18555a16247cfa4b64f41c74eebab19e4b4d18579c2612c1495d5R57) yesterday. I don't think they will conflict with each other when merged, but leaving a comment just in case.

PR #283, opened 2021-10-20
an initial react-datetime setup

luchux:
Done: 
- install react-datetime package. It provides a UI input element with a complete API to manage date in other components (as forms, etc. ) https://www.npmjs.com/package/react-datetime
- A simple InputDate raw component that renders an input with a calendar to pick from. 

Usage: 
- wherever we need input dates, we should use a <Datetime > with value as prop for controlled forms. 
- we can wrap it in a StyledComponent based on the  use case when needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "react": "^16.10.2",
     "react-app-rewired": "^2.1.6",
     "react-countdown": "^2.3.2",
+    "react-datetime": "^3.1.1",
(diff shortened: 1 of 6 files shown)


mprasanjith (changes requested):
I think you meant to open this to develop?

rossneilson (approved):

mprasanjith (approved):

PR #290, opened 2021-10-21
Feature/250 ether input

luchux:
Done: 
- Added YUP for validation schemas. 
- Added form validation only to RewardContracts form at the moment. 

ToDo after review: 
- Add for other forms. 
- See if we need extended schemas, and if we can reuse some according to similar forms. 

Actual rules: 
beneficiary: check non empty (we can do ether address verification or things like that if needed)
repChange: check non empty. check that is bigger or equal to 0 (= because the placeholder was set to 0). check that is not higher than 10 (just as an example if we need upper limits) 
ethValue: check non empty. check that is bigger or equal than 0. 
externalToken: check non empty. check that is bigger or equal than 0. 
tokenValue: check non empty. check that is bigger or equal than 0. check is not bigger than 9999. 

in all those fail checks, show the errors as in strings of yup. 
    
Some examples: 
Screen Shot 2021-10-21 at 4 45 28 AM
Screen Shot 2021-10-21 at 4 42 29 AM
Screen Shot 2021-10-21 at 4 45 15 AM

Screen Shot 2021-10-21 at 4 45 40 AM
Screen Shot 2021-10-21 at 4 49 08 AM


rossneilson (commented):
I havent used yup before but it seems like a good way of validating this.
Is it possible to add it also for the other call types for example all the calls supported by the quick wallet schemes etc?

luchux:
@rossneilson sure! lets chat about it.

src/pages/NewProposal.tsx
```diff
@@ -353,21 +385,15 @@ const NewProposalPage = observer(() => {
...
           }
         });
 
         value = calls.map(call => {
-          return callToController
-            ? '0'
-            : call.callType === 'simple'
-            ? library.utils.toWei(call.value).toString()
-            : call.value;
+          return callToController ? '0' : call.value;
```
luchux: @AugustoL @rossneilson since we are using eth instead of wei, this should not cast, right?

src/pages/NewProposal.tsx
```diff
@@ -353,21 +385,15 @@ const NewProposalPage = observer(() => {
             return daoService.encodeControllerGenericCall(
               call.to,
               callData,
-              call.callType === 'simple'
-                ? library.utils.toWei(call.value).toString()
-                : call.value
+              call.value
```
luchux: @AugustoL @rossneilson since we are using eth instead of wei, this should not cast, right?

luchux:
@rossneilson you can see how i added more schemes, and added minimum to some fields in calls. 

If we want all the fields marked we can do this as well. 
I will give a retouch to the styles of the forms still (error color border in input as well), and disable push to ifps  butotn if there are errors. 

So if you wanna discuss the rules for addresses, and values and messages, would be epic ;) Also 0 values, placeholder are 0, so can we send 0 value? same for SchemeRegister...

src/pages/NewProposal.tsx
rossneilson: I think this should be getting transformed to wei before we encode it right?
Ether is the human readable version but contracts want wei and so it needs the toWei function

rossneilson:
I may have missed something but I think these are still only being applied to the old daostack calls (Majority of schemes in mainnet and xdai). However since we are moving towards the new architecture and schemas it would be nice for this to work there also.
See the below screenshot for an example of where it would be nice to apply the validation to:
スクリーンショット 2021-10-28 13 25 31

src/pages/NewProposal.tsx
luchux: ok!

rossneilson (approved):

PR #291, opened 2021-10-21
feat: Make proposals list titles right-clickable.

### LINKED ISSUE #233: Enable right clicking on proposals in the home page list, so you can open a few proposals in new tabs.
If you sort the list of proposals and then want to open a few in new tabs, you can't. You have to enter each proposal one by one and when you come out of a proposal the sorting is lost. It would be cool to be able to right click on proposals in the list, so you can open them in new tabs and switch between them easily.

mprasanjith:
Fixes #233.

This is surprisingly tough to fix. Our proposals list uses html `table` / `tr` / `td` tags. To make a whole row right-clickable as a link, we have to have an anchor tag in the row level, which seems to be impossible with HTML tables.

What I finally did was just making the proposal title a clickable link, which is not perfect as the table highlight and the proposal title highlight are two different things.

I think its better to move to a CSS table / flex / grid at some point anyway as HTML tables have their limits.

### DIFF
--- a/src/pages/Proposals.tsx
+++ b/src/pages/Proposals.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect } from 'react';
 import styled from 'styled-components';
-import { useHistory } from 'react-router-dom';
+import { useHistory, Link } from 'react-router-dom';
 import { observer } from 'mobx-react';
 import MiniSearch from 'minisearch';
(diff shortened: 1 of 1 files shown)


rossneilson (approved):
Looks good for now, nice job.
Would be good in future the entire row could be right clicked. Also having ctrl + click work would be awesome.

PR #300, opened 2021-10-22
Refactor IPFS connectivity to handle initialization gracefully.

### LINKED ISSUE #198: Fix get of undefined IPFS service error
Running locally on develop branch there is an issue with ipfs service that tries to get ipfs data before starting the service.

It seems to happen more consistently if you reload a /proposal/id page as it immediately fetches ipfs data.

This should be fixed and also likely adding a try catch would be best.

mprasanjith:
Previously, there was a race condition that prevented IPFS content from being loaded when the IPFS client still had not initialized. This PR refactors the `IPFSService` to handle connectivity more gracefully.

As a consequence of this change, we no longer need to initialize the IPFS client on application startup. Instead, it will be initialized only when needed.

Fixes #198.

### DIFF
--- a/src/PageRouter.tsx
+++ b/src/PageRouter.tsx
@@ -42,7 +42,6 @@ const PageRouter = observer(({ children }) => {
       notificationStore,
       providerStore,
       configStore,
-      ipfsService,
       etherscanService,
       pinataService,
       coingeckoService,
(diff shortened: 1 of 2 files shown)


hhamud (approved):
LGTM!. Great Work

rossneilson (approved):
Looks great, initial uploading isnt even much slower

Issue #299, opened 2021-10-21
Keep search and filtering when navigating between results.


mprasanjith:
Search / filtering results get cleared when you open a proposal and then return back to the proposals list. These parameters and search results should be restored when coming back to the proposal list instead. We can even look into [restoring the scroll position](https://reactrouter.com/web/guides/scroll-restoration).

luchux:
Seems our best choice based on the actual patterns in the app, is to use
localStorage.

luchux:
PR opened, with query params

PR #303, opened 2021-10-26
Filter saved with queryparams

luchux:
Done:  https://github.com/DXgovernance/dxvote/issues/299 Keep filter in query params.


luchux:
what do you guys think of this navigation ? those filters changing themselves is like the app has the control haha. 
but if you gimme the ok with this, i will move the loading component to be reusable and could be ready to send.

rossneilson (approved):
Works great, nice job Lulli!

PR #304, opened 2021-10-26
Improved Wallet and Network Switching

### LINKED ISSUE #192: Prevent reload on network/chain change
Currently we fully reload the page on chainChanged in Web3ReactManager.
It seems this is an alright solution that is not uncommon, but if you look at something like swapr switching chains is a lot smoother and faster.

Copying how swapr does this is likely the easiest method to quickly switch chains without reloading the page.

mprasanjith:
Part 2 of my Web3 refactor. Please review and merge #301 before this.

- This changes the Web3 behavior to handle network and wallet switching without a reload. This required quite a bit of changes to clear the data in our MobX stores.
- I've also simplified and cleaned up the code for network handling in URLs.

Changes have been tested when disconnected from wallet, and with MetaMask and WalletConnect.

Fixes #192.

### DIFF
--- a/src/PageRouter.tsx
+++ b/src/PageRouter.tsx
@@ -55,7 +55,7 @@ const PageRouter = observer(({ children }) => {
   const { active: providerActive } = useWeb3React();
 
   useEffect(() => {
-    if (location.pathname == '/') {
+    if (location.pathname == '/' && networkName) {
       history.push(`/${networkName}/proposals`);
     }
   }, [networkName]);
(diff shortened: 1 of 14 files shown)


hhamud (approved):
Great Improvement, totally feel the difference. Code LGTM!

rossneilson:
Awesome work, super smooth!
However it looks like directly linking to a proposal on a different network to metamask isnt working anymore

mprasanjith:
@rossneilson Good catch. Didn't expect that to break. Will fix and update this PR. Making this a draft until then.

rossneilson (approved):

PR #319, opened 2021-10-29
Feature/update cache without release

### LINKED ISSUE #309: Store cache IPFS addresses in cache.dxvote.ens
**Description**
We want a way to update the cache without having to do a release, for this we will store a .json file in the content record for the cache.dxvote.ens domain.

**Proposed Solution**
1. The first place that the dxvote dapp should look for the default cache file is the cache.dxvote.ens content hash, this will return an IPFS hash and the file can be fetched from there, the json file will be `{[networkName]: [IPFS_Hash]}`.
2. For this we will need to make a call using the default rpc provider **in mainnet** no matter what the user network is, so we will likely need to instantiate a web3 object using the default RPC for mainnet and do the call to the ENS registry to fetch the IPFS hash and then fetch the hash.
3. If the process in the step 2 fails we will catch the error and get the cache from the default IPFS hash stored in the build.

**Expected Behavior**
The cache would be able to be updated by setting the cache.dxvote.ens domain, removing the need to release a build for it. BUT we will also update this "backup default cache" everytime we do a release.

mprasanjith:
Fixes #309.

We want a way to update the cache without having to do a release. To do this, the cache IPFS hashes would be set at cache.dxvote.eth. The app reads this metadata file from IPFS and resolves the correct content hash for the active network. If there's an issue, the app falls back to using the content hash in the configs.

PS: With the latest changes, this will now load the whole application config from ENS, not just the cache. It will fallback to the configs in the build on failure.


mprasanjith:
@AugustoL To test this, we'll need cache.dxvote.eth configured. Would you be able to set it up?

mprasanjith:
With the latest changes, this will now load the whole application config from ENS, not just the cache. It will fallback to the configs in the build on failure.

rossneilson (approved):

PR #333, opened 2021-11-02
feat: Import proposals from DAOtalk.

mprasanjith:


luchux (approved):
nice job

rossneilson (approved):
I got Eylon to add develop.dxvote.dev to a setting which may bypass cors so worth merging this to develop and testing it

PR #336, opened 2021-11-02
Refactor Proposal into subcoments

luchux:
Done: 
moved History, Votes, Details, Description, Stakes, Status as subcomponents. 

At the moment the naming is related to components/Proposal/ and each components is named as in: 
components/Proposal/Votes/index.tsx 

this give us space to move individual styles, tests, types, into 
components/Proposal/Votes/styles.ts
components/Proposal/Votes/tests.ts 
components/Proposal/Votes/types.ts 
etc. 

for now we just moved the code of the component and mainly all take the reusable styles from Proposal/styles.ts and implement their own in the ComponentName/index.ts

### DIFF
--- a/src/components/Proposal/Stakes/index.tsx
+++ b/src/components/Proposal/Stakes/index.tsx
@@ -1,15 +1,8 @@
 import { useState } from 'react';
-import styled from 'styled-components';
 import moment from 'moment';
-import {
-  Question,
-  AmountBadge,
-  BlockchainLink,
-  HorizontalSeparator,
-} from 'components/common';
+import { useLocation } from 'react-router-dom';
 
 import { useContext } from 'contexts';
-import { useLocation } from 'react-router-dom';
 import {
   bnum,
   calculateStakes,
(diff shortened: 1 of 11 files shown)


src/components/Proposal/Stakes/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+const Stakes = () => {
+  const {
+    context: { daoStore, configStore, providerStore, daoService, userStore },
+  } = useContext();
+
+  const [stakeAmount, setStakeAmount] = useState(0);
+  // We should get the ID in another way
+  const proposalId = useLocation().pathname.split('/')[3];
+  console.log('Proposal ID in Vote: ', proposalId);
```
hhamud, line 42: console.log should be removed

src/components/Proposal/Stakes/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  const { account } = providerStore.getActiveWeb3React();
+  const scheme = daoStore.getScheme(proposal.scheme);
+
+  let stakedAmount = bnum(0);
+  let positiveStakesCount = proposalEvents.stakes.filter(
+    stake => stake.vote.toString() === '1'
+  ).length;
+  let negativeStakesCount = proposalEvents.stakes.filter(
+    stake => stake.vote.toString() === '2'
+  ).length;
```
hhamud, line 54: I think we should talk to @AugustoL to find out why stakes.vote === 1 is positive and 2 is negative and if we can change this for clarity

src/components/Proposal/Stakes/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+    } else if (stake.staker === account && stake.vote.toString() === '2') {
+      stakedAmount = stakedAmount.minus(stake.amount);
+    }
+  });
+
+  const { recommendedStakeToBoost, recommendedStakeToUnBoost } =
+    calculateStakes(
+      votingParameters.thresholdConst,
+      scheme.boostedProposals,
+      proposal.stateInVotingMachine === 4
```
hhamud, line 84: I think we should stop comparing enums to numbers as its not clear what is happening here. thoughts?

src/components/Proposal/Votes/index.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  return (
+    <>
+      <SidebarRow>
+        <strong>
+          Votes <Question question="4" />
+        </strong>
+      </SidebarRow>
+      <SidebarRow>
+        <PositiveSummary>
```
hhamud, line 114: Positive and negative summary components here are very similar. Create a more generic component so that it can be reused here and cuts down on LOC

src/components/Proposal/Status/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        {status === 'Pending Execution' && executionTimeoutTime.toNumber() > 0 && (
+          <span className="timeText">
+            {' '}
+            Execution timeout in{' '}
+            <Countdown date={executionTimeoutTime.toNumber() * 1000} />{' '}
+          </span>
+        )}
+      </SidebarRow>
+      {account && (
+        <SidebarRow style={{ flexDirection: 'column', alignItems: 'center' }}>
```
hhamud, line 88: I think we can probably refactor this but I have no idea what `pendingAction` is supposed to represent?. @AugustoL

src/pages/Proposal.tsx
```diff
@@ -487,512 +147,21 @@ const ProposalPage = observer(() => {
...
-            }}
-          >
-            <ActionButton color="blue" onClick={() => redeemDaoBounty()}>
-              Redeem Stake Bounty
-            </ActionButton>
-          </SidebarRow>
-        ) : (
-          <div></div>
-        )}
+        <Stakes />
```
hhamud, line 164: nice! only 200 LOC here.

hhamud (approved):
LGTM overall but I left a few comments

src/components/Proposal/Status/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <FiPlayCircle /> Redeem 4 Beneficiary{' '}
+            </ActionButton>
+          ) : (
+            pendingAction === 5 && (
+              <ActionButton color="blue" onClick={executeMulticall}>
+                <FiPlayCircle /> Execute Multicall{' '}
+              </ActionButton>
+            )
+          )}
+        </SidebarRow>
```
hhamud, line 112: I think we can probably refactor this but I have no idea what `pendingAction` is supposed to represent?. @AugustoL

src/components/Proposal/Stakes/index.tsx
AugustoL: The voting machine was design to have multiple options, but we use two by default, yes and not, 1 and 2, we should compare this against an enum like `Vote.YES` instead of `"1"`. Something to be added in the future.

src/components/Proposal/Stakes/index.tsx
AugustoL: Yes, we should. Lets leave it for a future PR though,

PR #338, opened 2021-11-03
fix: Tracking abis properly per file and getting local controller abi

### LINKED ISSUE #302: Same function details being resolved when multiple calls use blockexplorer for ABIs
When a proposal has multiple calls that dont have defined strings for displaying call actions and instead use block explorer APIs to fetch ABIs there is an issue where all function names and params show up as the same.

スクリーンショット 2021-10-25 17 42 30

https://dxvote-dev.ipns.dweb.link/#/xdai/proposal/0xaeee624c3a249ae7059eee458cf1c00e58d998a59446c2e0b9b387b49901e81c

It seems the useAbiService is only storing one ABI at a time so it overwrites the ABI being used to the last call, in this case the create() function
Likely it should store a map indexed by address?

rossneilson:
Fixing duplicate call data appearing on more complex wallet scheme proposals.

Issues fixed:
- Controller was not verified so ABI was not found - Added check for the controller and pulling ABI locally if matched
- ABI value was storing single value  - Storing ABI in decoded calls object instead which is in an array

Closes #302

### DIFF
--- a/src/components/CallDataInformation.tsx
+++ b/src/components/CallDataInformation.tsx
@@ -48,7 +48,6 @@ const CallDataInformation = observer(
         proposal.to.map(item => getContractABI(item))
       );
       result.map((abi, i) => {
-        console.log({ abi });
         proposalCallArray.push(
           decodedCallData(
             scheme.type === 'WalletScheme' &&
@@ -62,7 +61,6 @@ const CallDataInformation = observer(
           )
         );
       });
-      console.log({ proposalCallArray });
       setLoading(false);
     };
     useEffect(() => {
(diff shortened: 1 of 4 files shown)


rossneilson:
All calls showing up as expected on xdai proposal 0xddc6e9530cc73fefaf65d57f5151fefc31ea9c5ea74b74487e75ddfead19971a

Screenshot 2021-11-03 at 13 03 45

hhamud (approved):
LGTM

hhamud:
> Fixing duplicate call data appearing on more complex wallet scheme proposals.
> 
> Issues fixed:
> 
> * Controller was not verified so ABI was not found - Added check for the controller and pulling ABI locally if matched
> * ABI value was storing single value  - Storing ABI in decoded calls object instead which is in an array
> 
> Closes #302

I heard about how verifying contracts on blockscout is a real pain.

PR #339, opened 2021-11-03
feat: Override stablecoin amount in contributor proposals.

### LINKED ISSUE #334: Add stablecoin override to contributor proposals
Currently the new contributor proposals screen covers the most basic use cases but some overrides in the advanced option would be useful for edge cases such as people requesting additional stable coins to cover travel expenses.

Ideally, this would look like an additional input available after clicking "More options" where the selected amounts of stables are visible but they can be edited to override the default level value.

Back on the main screen, if edited, the override should show instead of the default.

If the input is deleted and left blank then the default level value is used. 

In future we could also have advanced options specifically for travel expenses but for now this is more flexible and simple.

mprasanjith:
This adds support to override the requested stablecoin amount when submitting a contributor proposal. 

Fixes #334.

image
image

### DIFF
--- a/src/pages/ProposalSubmission/ContributorProposal.tsx
+++ b/src/pages/ProposalSubmission/ContributorProposal.tsx
@@ -92,6 +92,7 @@ const InputWrapper = styled.div`
   display: flex;
   justify-content: center;
   font-size: xx-large;
+  align-items: center;
 `;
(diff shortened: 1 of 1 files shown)


rossneilson (approved):

luchux (commented):
looks allright to me

PR #353, opened 2021-11-05
Refactoring Calls and Proposal styles

luchux:
Calls was hard to read. Splitted into subcomponents and touches some styles.


rossneilson:
Something I wonder if we could change in here.

I don't love the inconsistency of these lines, they are used to separate calls as well as sections of calls if it comes from etherscan.

I think it should only be separating calls

Example here https://bafybeiaapbbdofd6jszdhw3ayjbf3o5n7tgyksj6ycibs6wjfurr7druva.ipfs.dweb.link/#/xdai/proposal/0xddc6e9530cc73fefaf65d57f5151fefc31ea9c5ea74b74487e75ddfead19971a

Screenshot 2021-11-08 at 19 06 24

rossneilson (approved):

PR #359, opened 2021-11-08
warning of Dev Mode

luchux:
A way to show that we are not in production.

Screen Shot 2021-11-08 at 9 42 53 AM

### DIFF
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -75,6 +75,11 @@ const Header = observer(() => {
 
   const { active, account } = providerStore.getActiveWeb3React();
 
+  const isTestingEnv = !(
+    window?.location?.href?.includes('dxvote.eth.link') ||
+    window?.location?.href?.includes('dxvote.eth.limo')
+  );
+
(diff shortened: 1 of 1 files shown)


rossneilson:
I think the idea is more to show this message on any URL that is not dxvote.eth
So as a test, this message should appear on the ipfs deployment previews and only not show up on dxvote.eth.link | dxvote.eth.limo

Since we have a lot of QA testing environments that use the same environment config we need to show this there since anyone can access it.

Also I wonder if once it shows on those sites then maybe "Testing Environment" would be a better message

luchux:
@rossneilson changed it! checking for the domain safely with windows  (react-router dom useLocation is domain agnostic).

rossneilson:
> @rossneilson changed it! checking for the domain safely with windows (react-router dom useLocation is domain agnostic).

Awesome, we could even change it to just check for 'dxvote.eth'

luchux:
@rossneilson ok, np updated it.

rossneilson (approved):

PR #360, opened 2021-11-08
Feature/blockchainlink component updates

### LINKED ISSUE #355: Add ENS resolution back to header
**User Story**
As a user with an ENS, I want to see my .eth in the header so that I can verify the correct account easier

**Possible Solutions**
This was added in #295 but later removed due to the confusion that having a link in a button created (not noticed at time)
So the best solution may be a flag on BlockchainLink to allow the disabling of the clickable link in the text

mprasanjith:
This has a few refactors and bugfixes regarding the `BlockchainLink` component and ENS reverse resolution.

1. `BlockchainLink` component now has a `onlyText` prop, which enables to show just the resolved ENS name / wallet address, without the linking to the block explorer.
2. `BlockchainLink` has been added back to the application header, which fixes #355.
3. ENS reverse resolution has been moved to `ENSService`. Now this reuses the same `MainnetWeb3Provider` Web3 connection, without spawning a new one with ethers.js default settings.
4. ENS resolution errors now only throw a warning instead of an error.
5. `MainnetWeb3Provider` uses the RPC configs even when MetaMask is connected. The RPC configuration UI is now shown even when MetaMask is connected.

### DIFF
--- a/src/components/Web3ConnectStatus/index.tsx
+++ b/src/components/Web3ConnectStatus/index.tsx
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 import { observer } from 'mobx-react';
-import { toAddressStub, toCamelCaseString } from '../../utils';
+import { toCamelCaseString } from '../../utils';
 import WalletModal from 'components/WalletModal';
 import { getChains, injected, isChainIdSupported } from 'provider/connectors';
 import { useContext } from '../../contexts';
@@ -8,6 +8,7 @@ import { Box } from '../../components/common';
 import NetworkModal from 'components/NetworkModal';
 import { useEffect, useState } from 'react';
 import { useRpcUrls } from 'provider/providerHooks';
+import BlockchainLink from '../common/BlockchainLink';
 
 const WrongNetworkButton = styled(Box)`
   color: var(--dark-text-gray);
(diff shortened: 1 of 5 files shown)


rossneilson:
This isnt resolving any ENS names for me

mprasanjith:
> This isnt resolving any ENS names for me

Oh, missed this comment earlier. Seems dweb.link preview deploys are not whitelisted on Alchemy. This is causing the ENS resolution to fail.

The new change uses the RPC configured in Settings to resolve ENS names. (Even when connected to MetaMask) Just tested by adding my own Alchemy key, and it works.

rossneilson:
> The new change uses the RPC configured in Settings to resolve ENS names. (Even when connected to MetaMask) Just tested by adding my own Alchemy key, and it works.

Could we use the previous approach as a backup that doesn't rely on alchemy?

mprasanjith:
> Could we use the previous approach as a backup that doesn't rely on alchemy?

The previous method was to use the [default ethers.js provider](https://docs.ethers.io/v5/api/providers/#providers-getDefaultProvider), using default API keys shipped with Ethers.js. These keys are shared among all Ethers.js users, and I don't think its good for production use.

I guess what we can do is to have a development-only unrestricted Alchemy / Infura API key for our preview deployments, separate from production? This should also solve the issue of DXvote not loading in preview deploys until MetaMask is connected.

rossneilson:
Ah gotcha, nice work then

rossneilson (approved):

PR #361, opened 2021-11-09
Feature/guilds frontend setup

mprasanjith:


src/pages/Guilds/Guilds.tsx
```diff
@@ -0,0 +1,112 @@
...
+      <SidebarContent>
+        <Sidebar />
+      </SidebarContent>
+      <PageContent>
+        <PageHeader>
+          <Box>
+            <PageSubtitle>DXdao</PageSubtitle>
+            <PageTitle>Proposals</PageTitle>
+          </Box>
+          <DropdownMenu position={DropdownPosition.BottomRight}>
```
luchux, line 84: you could potentially move this filter to a Filter component, because otherwise this page is gonna mix all the filter event handlers and the proposal filtering. Doesn't seems to be much the responsibility of the page that imho.

src/components/Guilds/Sidebar/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+import dxIcon from '../../../assets/images/dxdao-icon.svg';
+
+const SidebarWrapper = styled(Box)`
+  border: 1px solid #000;
+  border-radius: 0.5rem;
+  margin-right: 1rem;
+`;
+
+const SidebarHeader = styled(Box)`
+  border-bottom: 1px solid #000;
```
luchux, line 15: colors and fonts maybe from theme?

src/components/Guilds/Header/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+const MenuItems = styled(Box)`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-left: auto;
+`;
+
+const AppLogo = styled(Box)`
```
luchux, line 26: given the design files, i don't think we are gonna be able to maintain well DxVote and Guilds simultaneously on the same Layout components, since they are very different and we will always be extending / overwritin properties. imho

rossneilson (approved):

src/pages/Guilds/Guilds.tsx
mprasanjith: Definitely. I was thinking of waiting until we start wiring up actual data. I'm right now working on the single proposal page, and once that's done I think we can look at how to structure everything better.

luchux (approved):

src/components/Guilds/Sidebar/index.tsx
mprasanjith: Yeah, definitely should implement a new theme, but was hesitant as the designs are still wireframes. I'll be putting a new PR this week with more changes, will implement theme stuff there. Thanks for the suggestion!

src/components/Guilds/Header/index.tsx
mprasanjith: DXvote is going to get a new UI design that should align with Guilds. I know it'd be a huge refactoring though.

rossneilson: Guilds has priority over design as it is the product we will be launching externally first but eventually DXvote will be ported over to use the same base components and stylings

PR #365, opened 2021-11-10
feat: Adding date input to contributor proposal

### LINKED ISSUE #342: Calendar select for worker proposal second period
Currently, in the contributor proposal creation UI there is a selection for the first period or the second period of the 2 month worker proposal (start/end)

Since payment should be taken from the initial proposal date we are currently doing some automated calculations for REP percentage (ContributorProposal.tsx - ln 188). This set bounds on dates of REP events for 1-3 months and looks for the largest REP event for that user, it then checks if the REP amount is roughly similar to the expected inflation rate to decide if its a correct REP amount.

As much as I am pretty proud of this solution and how it seems to work quite well ... its a little overly complex for a small amount of automation.

The better solution would simply be showing a calendar input for the user to select the proposal starting date if this is the second / end proposal for the period. From here the correct REP amount can be calculated.

This also allows us to adjust the ATH amount for DXD calculation, if a new ATH was hit since the first proposal date then the previous ATH should be used for payout calculation.

rossneilson:
This allows users to set the date to correctly calculate DXD price and REP allocation

For DXD if the start period is after the most recent ATH, then ATH is used. If before ATH then the user is prompted for the DXD price at time of proposal creation. (Wasnt possible to easily automate with coingecko API)

getRepAt function has been updated with optional atTime parameter allowing for the specification of a timestamp instead of block number which is used to find the total supply at the specified date to correctly calculate REP award. 

This contributes more to the problem of this page growing and needing split up which I believe is covered in the refactor. Sorry for adding more work to it. 

Closes #342 

Screenshot 2021-11-10 at 11 37 47

### DIFF
--- a/src/components/common/Forms/InputDate.tsx
+++ b/src/components/common/Forms/InputDate.tsx
@@ -26,7 +26,8 @@ const Date = styled(Datetime)`
     width: ${props => `${props.width}px`};
   }
   .rdtPicker {
-    width: 100%;
+    left: 50%;
+    transform: translateX(-50%);
   }
 `;
 
--- a/src/components/common/Forms/InputDate.tsx
+++ b/src/components/common/Forms/InputDate.tsx
@@ -32,7 +32,7 @@ const Date = styled(Datetime)`
 
 export const InputDate = ({ onChange, value, text = null, width = 200 }) => (
   <div>
-    <Label>{text}</Label>
+    <Label width={width}>{text}</Label>
     <Date
       value={value}
       onChange={onChange}
--- a/src/components/common/Forms/InputDate.tsx
+++ b/src/components/common/Forms/InputDate.tsx
@@ -1,4 +1,5 @@
 import Datetime from 'react-datetime';
+import styled from 'styled-components';
 import 'react-datetime/css/react-datetime.css';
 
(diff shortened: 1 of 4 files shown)


mprasanjith:
The positioning of the date picker is not right for me. Checked on both Edge and Firefox.

image

Are we removing the First / Second Proposal toggle?

rossneilson:
> The positioning of the date picker is not right for me. Checked on both Edge and Firefox.
> 
> Are we removing the First / Second Proposal toggle?

Oh, I had console open on right so never noticed :D
Will look into it.

I had originally kept the toggle but realised it didn't make much sense if the date input was there as it didn't do anything.

rossneilson:
Positioning the date picker is a bit of a pain since it uses absolute positioning so as not to affect surrounding components. But this makes it difficult to center, so width 100% it is. 

Screenshot 2021-11-12 at 10 56 44

mprasanjith:
I also gave it a shot and got it working on the browser devtools inspector with the following:

```css
.rdtPicker {
    position: absolute;
    left: 50%;
    transform: translateX(-50%);
}
```

Can you check if this fixes the positioning issue?

mprasanjith (approved):

PR #371, opened 2021-11-11
Feature/rep display component

### LINKED ISSUE #315: REP display component
Since REP is a high security part of DXvote it would be nice to have a REP component that we can use when displaying an amount of REP that also shows in brackets that amount as a percentage of current total supply.

It could also take a parameter for `date` that would display "Total supply % as of `date`"

This would be most useful when displaying call data for `mintReputation`

mprasanjith:
This PR adds a simple reusable `RepDisplay` component which calculates and shows the REP % when hovering. It can also take a block number and adjust the percentage value accordingly. If a timestamp is passed in, it will show the date with the REP percentage.

I've also refactored the `RecommendedCalls` component to show `RepDisplay` in proposal call data. After spending too much time trying to do this with the existing `replaceAll`  method we were using, I gave up and changed the code to use a tiny utility library. (Let me know if that's not okay) Hopefully, now the code is cleaner and accommodating for any future React components we want to mix with the call data details.

Regular State:
image

On Hover:
image

This change closes #315.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "react-scripts": "^4.0.3",
     "react-select": "^4.3.1",
     "react-spring": "^8.0.27",
+    "react-string-replace": "^0.4.4",
     "react-switch": "^6.0.0",
     "react-use-gesture": "^7.0.16",
     "request": "^2.88.2",
--- a/src/components/Proposal/Calls/RecommendedCalls.tsx
+++ b/src/components/Proposal/Calls/RecommendedCalls.tsx
@@ -1,6 +1,10 @@
-import { normalizeBalance } from 'utils';
+import reactStringReplace from 'react-string-replace';
+import { useLocation } from 'react-router-dom';
+import { bnum, normalizeBalance } from 'utils';
 import { BlockchainLink } from 'components/common';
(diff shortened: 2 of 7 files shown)


mprasanjith:
Let me know any other places this component should be implemented in. :)

rossneilson:
This is awesome, amazing work. Interesting seeing that some REP amounts are incorrectly calculated, once we get using the automated version that doesn't rely on a spreadsheet this will be much more accurate. 

Not sure if this is maybe too much work here, but likely shouldn't be.
Could we add the first call here to the recommendedCalls so that it also shows the REP component since its what we will be moving to for contributor proposals

https://bafybeig7tigu7icijmh2dmpizs5lcvmnxpgoqhokxax3lir7l3jee5ta3y.ipfs.dweb.link/#/xdai/proposal/0xddc6e9530cc73fefaf65d57f5151fefc31ea9c5ea74b74487e75ddfead19971a
Screenshot 2021-11-12 at 11 05 00

rossneilson:
Turned out to be an easy fix, added myself

rossneilson (approved):

PR #388, opened 2021-11-17
feat: Add vesting contract creation to recommended calls for clear di…

rossneilson:
With blockscout giving CORS errors even in production which previously worked, we need to display what contributor proposals are doing.

This adds the vesting contract factory with the method create

Screenshot 2021-11-17 at 15 25 31

As a side note we need to translate these into ens names as well as proper times instead of seconds.


hhamud (approved):
LGTM

PR #391, opened 2021-11-18
Revert "Feature/250 ether input"

rossneilson:
Reverts DXgovernance/dxvote#290

I tested this again in our deployed develop instance and it is still broken.

Lets revert this for the release and take a closer look at why strange behaviour is happening with inputs that have had validation added to them,


src/pages/ProposalSubmission/Custom.tsx
```diff
@@ -310,15 +302,21 @@ const NewProposalPage = observer(() => {
             return daoService.encodeControllerGenericCall(
               call.to,
               callData,
-              call.value
+              call.callType === 'simple'
```
hhamud, line 305: I would encapsulate this as a function, it's repeated at line 317 and it would help to avoid nested ternary operators

src/pages/ProposalSubmission/Custom.tsx
```diff
@@ -310,15 +302,21 @@ const NewProposalPage = observer(() => {
...
           } else {
             return callData;
           }
         });
 
         value = calls.map(call => {
-          return callToController ? '0' : call.value;
+          return callToController
+            ? '0'
+            : call.callType === 'simple'
```
hhamud, line 317: mentioned previously

hhamud (approved):

src/pages/ProposalSubmission/Custom.tsx
rossneilson: Absolutely, not my code though, just a revert :P

PR #392, opened 2021-11-18
Feature/fix sorting executed

### LINKED ISSUE #307: Proposals ordering issues
It seems proposals are backwards on mainnet under "Execution successful" and so beyond currently active proposals what is shown is the very first proposals in DXdao.
Also currently Expired in Queue is too far up the list. In networks like rinkeby where it happens a lot this is very annoying since new proposals appear very far down the page. Would be great to treat Expired in queue the same as executed.
スクリーンショット 2021-10-26 21 34 33

rossneilson:
Closes #307 

A simple fix for proposals ordering since we want to see the most recent finish times on executed proposals not the earliest ones first as we do with in progress ones

### DIFF
--- a/src/utils/proposals.ts
+++ b/src/utils/proposals.ts
@@ -355,6 +355,9 @@ export const decodeProposalStatus = function (
 export const orderByNewestTimeToFinish = (a: any, b: any) =>
   a.finishTime - b.finishTime;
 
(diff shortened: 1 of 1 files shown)


hhamud (approved):
LGTM

Issue #399, opened 2021-11-18
Export recent proposals to markdown format


powerschris:
Definitely a *power user* request, but when Dave and I put together the Governance discussion agenda every week, we go over proposals that are 'Boosted' or pending boosting and we post links to those proposals in the forum. It would be extremely helpful if we could create that output through DXvote somehow. 

Just a list with the proposal title and link to that proposal would be enough. And markdown is preferable, so that it could be pasted easily into DAOtalk. 
You can see an example agenda here:
https://daotalk.org/t/governance-discussion-nov-17-2021/3549 
<img width="792" alt="Screen Shot 2021-11-18 at 10 23 07 AM" src="https://user-images.githubusercontent.com/47428627/142444102-f9c4b278-b4c7-46ca-90e8-ca0734d187a0.png">

**Alternative Solutions**

**Additional context**

PR #402, opened 2021-11-19
Feature/export data

### LINKED ISSUE #364: Export as CSV
One of my favorite features across the internet is the ability to take the data I'm looking at and export it to CSV format, so it can be analyzed/charted/combined with other things. Coingecko has this feature for their market data, but I think Etherscan does it best. They have a 'Download as CSV' button at the bottom of every page with transaction info, and then allows you to select the dates you want the data. Like so:

Screen Shot 2021-11-09 at 5 39 59 PM

Screen Shot 2021-11-09 at 5 40 04 PM

I think this would be very cool to add to different pages of DXvote. The three that I thought of:
1. Export all proposals as CSV, just a button on the bottom of the proposal view page that would allow you to export over certain time period. Would need to include proposal time/execution, proposer, proposal ID, scheme and any funds transferred or REP minted
2. Finance tab - export the assets in DXdao's treasury into CSV
3. REP holder profile page - would love to export my DXvote transaction history to CSV. When was I most active? How often did I stake. This could actually be very helpful for administrative/tax purposes.

**Alternative Solutions**

**Additional context**

mprasanjith:
Adds features to export data to different formats:
- Proposals list can be exported as CSV or copied as Markdown.
- Assets list in the finances tab can be exported as CSV.
- REP holder profile events can be exported as CSV.

When exporting proposals, a date range can be selected from the UI optionally as well.

image

I had to add some spacing below the popup content, as our date picker and the modal wasn't working well together. Hope we can sort it out at some point.

The `useExporters` hook is reusable, and can be used to export any array of JS objects as a CSV. We can reuse this hook in the future when we want to export different file formats / data.

This fixes #364 and #399. Exported CSVs at the moment might be missing some fields that were requested (any funds transferred or REP minted for a proposal for example). These values aren't currently available in the proposals page and might be too expensive to calculate.


rossneilson:
Would be good to get Chris' feedback on this but looks like preview got blocked by code format checks, could you make them pass?
Also maybe worth automating it in your IDE

rossneilson (dismissed):

rossneilson (approved):

PR #404, opened 2021-11-22
feat: Added pokt urls and service

### LINKED ISSUE #390: Add pokt RPC as default in QA environments
Pokt is a decentralised RPC that we are planning on supporting and partnering with.

Would be good to test it as default RPC on QA environments since we rely so heavily on RPCs in DXvote with no graph.

Supports mainnet and xdai

https://eth-mainnet.gateway.pokt.network/v1/lb/610e4d3118656e00369b9b05
https://poa-xdai.gateway.pokt.network/v1/lb/610e4ed818656e00369b9ccc

rossneilson:
Closes #390 

Requires a kind of ugly fix to force RPC connections to use correct header content types as there is no way of altering them with web3-react and pokt returns a 415 if incorrect.

Uses pokt as default RPC for mainnet, rinkeby and xdai

### DIFF
--- a/src/components/Web3ReactManager/index.tsx
+++ b/src/components/Web3ReactManager/index.tsx
@@ -20,7 +20,6 @@ const Web3ReactManager = ({ children }) => {
   const location = useLocation();
   const history = useHistory();
   const rpcUrls = useRpcUrls();
-  console.log({ rpcUrls });
 
   const originalFetch = window.fetch;
   window.fetch = (url, opts): Promise<Response> => {
--- a/src/components/Web3ReactManager/index.tsx
+++ b/src/components/Web3ReactManager/index.tsx
@@ -21,6 +21,7 @@ const Web3ReactManager = ({ children }) => {
   const history = useHistory();
   const rpcUrls = useRpcUrls();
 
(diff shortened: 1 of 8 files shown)


src/components/Web3ReactManager/index.tsx
```diff
@@ -21,6 +21,16 @@ const Web3ReactManager = ({ children }) => {
   const history = useHistory();
   const rpcUrls = useRpcUrls();
 
+  const originalFetch = window.fetch;
```
hhamud, line 25: Can you add a comment here on what it does?. The whole chunk of code here

rossneilson: Done, good call

rossneilson:
Just noticed we need the fetch override for things like ENS also, need to add to better location

rossneilson:
Added rpcUrls to effect on blockchain link that updates ENS names once rpcUrls are set since this means the native window fetch has been overridden therefor making pokt RPC connections for mainnet ENS resolution not return a 415 as they were doing before

hhamud (approved):

PR #418, opened 2021-11-23
Bugfix/duplicate events

luchux:
- A hot fix to avoid adding duplicate Votes, Stake and Redeem to the network cache. 
- A log for when a duplication happens, showing the event so we can trace back why this is happening.

### DIFF
--- a/src/stores/BlockchainStore.ts
+++ b/src/stores/BlockchainStore.ts
@@ -224,7 +224,7 @@ export default class BlockchainStore {
             chainId
           );
 
-          const fromBlock = lastCheckedBlockNumber;
+          const fromBlock = lastCheckedBlockNumber + 1;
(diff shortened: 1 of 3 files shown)


AugustoL (approved):

src/stores/BlockchainStore.ts
```diff
@@ -224,7 +224,7 @@ export default class BlockchainStore {
             chainId
           );
 
-          const fromBlock = lastCheckedBlockNumber;
+          const fromBlock = lastCheckedBlockNumber + 1;
```
rossneilson, line 227: I actually wonder if this alone might fix it, do you still see duplication logs after this change?

mprasanjith (approved):

rossneilson (approved):

PR #420, opened 2021-11-24
Feature/guilds proposal page UI

mprasanjith:
This adds the Proposal page UI base implementation. This is far from done, but let's get this merged as this PR is already quite big.

image


src/components/Guilds/common/Button/Button.tsx
```diff
@@ -1,31 +1,54 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
```
rossneilson, line 1: Is css needed?

rossneilson (approved):

PR #425, opened 2021-11-25
Feature/calls simple advanced toggle

tiagofneto:
Submission calls simple and advanced options button changed to switch

screenshot-localhost_3000-2021 11 25-13_34_06


rossneilson:
Hey, my bad for not adding this to the description of the task.

There is an existing component (src/components.toggle.tsx) that would be good to use here.
Only change needed would be some params for height and width (set default values to the current ones hardcoded in there)

rossneilson (approved):

PR #432, opened 2021-11-26
feat: Add wallet and network switching to guilds.

### LINKED ISSUE #424: Guilds web 3 provider
We have a pretty good existing implementation of web3 providers and wallet switching etc.
We should be good to just drag and drop that over to guilds. If any DXvote specific code exists it should be extracted.
Styling can be redone once we get designs for it.

mprasanjith:
The Web3Provider from DXvote alongside its wallet and network switcher UIs has been reused to add wallet and network switching support to Guilds. We'll need to update these UIs once we have new UI mockups for these components.

The wallet connections are shared between the two apps, so if the wallet is connected to DXvote, it would appear in Guilds and vice-versa. To support this I had to keep using MobX store.

Fixes #424.


mprasanjith:
PS: There is a redirect happening when switching networks. This is coming from the DXvote network switcher UI. I'm not going to touch it as anyway we'll replace this UI soon.

hhamud:
> PS: There is a redirect happening when switching networks. This is coming from the DXvote network switcher UI. I'm not going to touch it as anyway we'll replace this UI soon.

Can you make an issue of this when you can so that other's will know about this issue after this PR is merged.

hhamud:
I'm also getting an error when I try and connect my metamask to this. 
image

mprasanjith:
@hhamud Interesting. I was trying to reproduce this issue, but couldn't. Can you provide steps on what you did?

hhamud:
> @hhamud Interesting. I was trying to reproduce this issue, but couldn't. Can you provide steps on what you did?

upon re-testing this, its a general issue with the wallet connector when one connects to the application for the first time and clicks away from the popup of the wallet without accepting and then attempts to connect again.

src/components/Guilds/Header/WalletButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { useEffect, useState } from 'react';
+import { useRpcUrls } from 'provider/providerHooks';
+import { useContext } from '../../../contexts';
+import { Button, IconButton } from '../common/Button';
+import dxIcon from '../../../assets/images/dxdao-icon.svg';
+
+const ButtonIcon = styled.img`
+  height: 1.1rem;
+  width: 1.1rem;
+`;
```
hhamud, line 15: This component is repeated across WalletButton and NetworkButton, I suggest that you should create a common ButtonIcon component if it does not already exist

src/components/Guilds/Header/NetworkButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+import xdaiIcon from '../../../assets/images/xdai.svg';
+
+const iconsByChain = {
+  1: ethereumIcon,
+  4: ethereumIcon,
+  100: xdaiIcon,
+  42161: arbitrumIcon,
+  421611: arbitrumIcon,
+  1337: ethereumIcon,
+};
```
hhamud, line 20: Can't we place this in constants?

hhamud (changes requested):

src/components/Guilds/Header/NetworkButton.tsx
mprasanjith: This was in constants, but I think we moved this here as its better to only import it when we use it. (These icons are somewhat large.)

src/components/Guilds/Header/WalletButton.tsx
luchux: you can also create a base CSS and import that (instead of a component).

luchux (approved):

luchux (dismissed):

src/components/Guilds/Header/WalletButton.tsx
mprasanjith: This is addressed in my other PR (#449) which handles ENS avatar loading.

mprasanjith:
> > @hhamud Interesting. I was trying to reproduce this issue, but couldn't. Can you provide steps on what you did?
> 
> upon re-testing this, its a general issue with the wallet connector when one connects to the application for the first time and clicks away from the popup of the wallet without accepting and then attempts to connect again.

Would you be able to open an issue for this?

hhamud (dismissed):

rossneilson (approved):

PR #433, opened 2021-11-27
Filters Badges and logics context and hooks

luchux:
Adding some Filter Hooks and Context to maintain the state of the filter, and the counter of schemes and status selected. 
Adding the badges components
Adding the logic to show that badges with selected filter. 
https://watch.screencastify.com/v/YyfKlRzghHo6lNn1mldh


src/configs/proposalTitles.json
```diff
@@ -1,66 +1 @@
-{
-  "0x08382f2197b647b338a15bcb33287dc3edc278f8c089a12952b87ee3fbadb42a": "Update swapr.eth resolver to point to beta 8.1 release of Swapr",
-  "0x57985dd9a8a45009b86a82a20d3669e570106014aa4b1b46bbc96875977a5c28": "REP Transfer - fluiddrop - MINT proposal",
-  "0x83c84a67ac418e87269150ca8c947e9f77f1d474124c4de59dd53dc251617059": "REP Transfer - fluiddrop - BURN proposal",
```
luchux: @rossneilson not sure why this config files went through. Should remove them? or this is generate again later?

src/configs/localhost/config.json
```diff
@@ -1,8 +1,8 @@
 {
   "cache": {
     "fromBlock": 0,
-    "toBlock": 76,
```
luchux: @rossneilson same with this, shouldn't probably we here ey?

src/configs/proposalTitles.json
rossneilson: Its generated when we build the cache but best option here is keeping it as it was before

hhamud: `git add -p` solves this issue.

hhamud: `git add -p` prevents similar issues occurring

src/contexts/Guilds/filters.tsx
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext } from 'react';
+import { useMenu } from '../../hooks/Guilds/useMenu';
+
+const FilterContext = createContext(null);
```
hhamud, line 4: I had similiar thoughts on doing this for dxvote as well. Maybe we do the same in the future to keep consistency?

rossneilson (approved):

PR #438, opened 2021-11-29
Feature/update cache hamza

hhamud:
* Hashes of my config files upon attempt at updating the cache
* is it worth setting up logging to help you guys identify why the hashes are not the same?.


rossneilson:
Hey, think we are covered from the other PR #434

PR #448, opened 2021-11-30
guilds proposal page mobile

luchux:
Done: 
- added the media queries for the mobile view of proposal page in Guilds. 
- modify some styles as in the figma files. 

screencapture-localhost-3000-2021-11-30-10_54_59

### DIFF
--- a/src/components/Guilds/ProposalStatus/index.tsx
+++ b/src/components/Guilds/ProposalStatus/index.tsx
@@ -1,10 +1,11 @@
 import styled from 'styled-components';
 import { Box } from '../common/Layout';
+
 const Status = styled.div`
   display: flex;
   flex-direction: row;
   width: 100%;
-  justify-content: flex-end;
+  justify-content: flex-start;
   align-items: center;
 `;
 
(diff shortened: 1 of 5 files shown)


src/components/Guilds/ProposalStatus/index.tsx
```diff
@@ -1,10 +1,11 @@
 import styled from 'styled-components';
 import { Box } from '../common/Layout';
+
 const Status = styled.div`
   display: flex;
   flex-direction: row;
   width: 100%;
-  justify-content: flex-end;
+  justify-content: flex-start;
```
hhamud: changing this value to start, is causing this issue on the proposals page, where the status button is squishing up the text
image

hhamud (changes requested):
The proposal looks great but for some reason the changes have affected the proposals page as well here.

src/components/Guilds/ProposalStatus/index.tsx
luchux: nice!

hhamud (changes requested):
You fixed the previous issue with the status button but now there's a new issue here that seems to have broken. 
image

hhamud (approved):

PR #449, opened 2021-12-01
Feature/guilds ENS address component

mprasanjith:
Resolves wallet addresses to their ENS addresses and looks up avatar assigned to the ENS addresses.

Avatar support is based on the spec available [here](https://gist.github.com/Arachnid/9db60bd75277969ee1689c8742b75182). This currently can resolve avatars set as HTTP / HTTPS and IPFS URIs, and even NFTs conformant to ERC-721 and ERC-1155 specs. In the future we can support additional protocols, if needed. (Arweave etc.) If its not possible to get the avatar image due to CORS issues, we fall-back gracefully to [ENS Metadata Service](https://metadata.ens.domains/docs).

image

image

In case no avatar or ENS address is present, the code falls-back to a generated avatar (the same one MetaMask shows) and the regular Ethereum wallet address.

image

The code uses a hooks-based approach inspired by Uniswap. All pieces of logic has its own hook, and should make things quite modular and reusable.


mprasanjith:
We need to merge #432 before merging this.

src/components/Guilds/Avatar/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+      const tokenUri = await erc1155Contract.methods.uri(tokenId).call();
+      const nftMetadataUri = resolveUri(tokenUri);
+      const nftMetadata = await resolveNFTMetadata(nftMetadataUri);
+      return resolveUri(nftMetadata.image);
+    }
+
+    return null;
+  }
+
+  const resolveUri = (avatarUri: string) => {
```
luchux: can we move some of this functions to a hook, or utils? I feel this will need to be reused out of this component but may be wrong

src/components/Guilds/Header/WalletButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }, [injected]);
+
+  const { chainId, account } = providerStore.getActiveWeb3React();
+  const { ensName, avatarUri } = useENSAddress(account);
+
+  const toggleWalletModal = () => {
+    modalStore.toggleWalletModal();
+  };
+
+  const switchNetwork = async chain => {
```
luchux: I think also the network methods of swithNetwork, and get actual network, and whatever is about the network could be in a hook. Because we may need to change networks from other places? may be im wrong too!

luchux (commented):
let some questions.

mprasanjith:
@luchux all good ideas! I'm going to refactor these into hooks. Will put this in draft mode until then.

rossneilson:
Found a solution to the CORS issue plaguing most ENS avatars
The brand new, still in alpha, ENS metadata service
https://metadata.ens.domains/docs
https://metadata.ens.domains/mainnet/avatar/rossdev.eth
This could maybe be the backup option since I imagine it is just a centralised solution
So if CORS, go here

rossneilson:
Zett pointed out it may be better if we make the content of the component, the ens avatar, larger and the same height as the border with little padding

Screenshot 2021-12-02 at 09 52 55
Screenshot 2021-12-02 at 14 24 09

rossneilson:
Chatted to @pimato about the size of the PFP and maybe best to leave it for now and we can revisit when implementing styles

src/components/Guilds/Header/WalletButton.tsx
mprasanjith: I'll handle a refactor of how we handle this in my next PR. Thinking of separating our Guilds web3 implementation from the DXvote one, so we can fix a bunch of issues.

rossneilson:
Found an issue in mobile
Screenshot 2021-12-07 at 17 01 22

rossneilson (approved):
Looks amazing and great refactor!

PR #478, opened 2021-12-08
[WIP] Find proposal Ids of guild

### LINKED ISSUE #464: Find proposal Ids of guild
This will be the first time we link up guild contracts to the UI 

The first step will be to query all the proposals of a certain guild, requiring a few steps:
- Setup ABI and helper files for ERC20Guild
- Add guild address to the url
- Call getName() and getToken() for guild
- Call getProposalIds on first load (https://github.com/DXgovernance/dxdao-contracts/blob/b7eb732c0c3c81ea53ed928a3077d71781941b50/contracts/erc20guild/ERC20Guild.sol#L833)
- Reverse sort and make available in UI

https://rinkeby.etherscan.io/address/0x9cDC16b5f95229b856cBA5F38095fD8E00f8edeF#readContract
Test on rinkeby
You may need to create an initial proposal to test also, see dxdao-contracts test code for examples

### LINKED ISSUE #465: Get proposal data from id
Blocked by #464

Once we have our array of proposal Ids we can begin fetching the state data to display
- Call getProposal(proposalId) of the current guild - https://github.com/DXgovernance/dxdao-contracts/blob/b7eb732c0c3c81ea53ed928a3077d71781941b50/contracts/erc20guild/ERC20Guild.sol#L850
- Do this for the first 5 proposals in the array
- Make proposal state data available  

Also create a typescript interface for this object

hhamud:
### Complete:
* Closes #464 
* Closes #465 
* Creates new `useProposals`hook in guilds
* Creates new `GuildProposals` interface
* refactored `useContract` hook
* Implement into guilds proposals page


src/abis/ERC20Guild.json
```diff
@@ -0,0 +1,1031 @@
+{
```
rossneilson: Do we need ABIs duplicated here as well as in contracts/ ?

src/configs/proposalTitles.json
```diff
@@ -1,54 +1,41 @@
 {
```
rossneilson: Could you remove these changes, we should keep this update to specifically cache update PRs

rossneilson:
Could we implement this on page load and pull the guild address from the URL?
currently its not used anywhere

src/configs/proposalTitles.json
hhamud: I was getting merge conflicts but I will try that

src/abis/ERC20Guild.json
hhamud: I'm not sure what the convention is since @mprasanjith created this folder. I can remove this file specifically and keep it in contracts/

hhamud:
> Could we implement this on page load and pull the guild address from the URL? currently its not used anywhere

It would be best to make a separate issue on making the proposals page interactive.

rossneilson:
> > Could we implement this on page load and pull the guild address from the URL? currently its not used anywhere
> 
> It would be best to make a separate issue on making the proposals page interactive.

I didn't mean interactivity, just calling it so we can test this works. Goal of these issues was to have the data available in the UIs

src/pages/Guilds/Guilds.tsx
```diff
@@ -47,20 +49,31 @@ const proposalsMock = [
 ];
 
 const GuildsPage: React.FC = () => {
+  // @TODO: parse guild contract address from where?
+  // @TODO: loading designs
+  // @TODO: error designs
+  const GUILDS_ADDRESS = '0x9cDC16b5f95229b856cBA5F38095fD8E00f8edeF';
```
rossneilson: Issue mentions pulling address from the URL

> Add guild address to the url
#464

mprasanjith: Can we move this to a constants file or a config? I think all chains will have the same contract address, but will need to confirm that.

src/pages/Guilds/Guilds.tsx
```diff
@@ -33,7 +35,7 @@ const ProposalsList = styled(Box)`
   margin-top: 1rem;
 `;
 
-const proposalsMock = [
+export const proposalsMock = [
```
mprasanjith: I think we can get rid of this now?

mprasanjith (dismissed):
Great work! Code looks very clean!

Couple more suggestions we can implement now or make issues for:
1. I feel like we should run `typechain` as a postinstall hook.
2. The loading spinner could be centered? I don't think we have any designs for this yet, so maybe this can wait.

rossneilson:
I created a test proposal and it caused the UI to crash, maybe a few things to iron out there

src/pages/Guilds/Guilds.tsx
mprasanjith: @rossneilson suggested using this from the URL, which is the proper way to do this. I think we can redirect from `/guilds` to `/guilds/0x9cDC16b5f95229b856cBA5F38095fD8E00f8edeF` for now?

rossneilson (changes requested):
Just marking this as needing changes to avoid the UI breaking if merged, see above comment

hhamud:
> I created a test proposal and it caused the UI to crash, maybe a few things to iron out there

Can you post exactly what error message you got here, because I know currently that when you try to refresh the guilds page, that it crashes and that you need to have a full refresh from the main page in dxvote because of the way that it is currently coupled.

src/pages/Guilds/Guilds.tsx
hhamud: What is the best way of fetching the guild's contract address?.

hhamud: https://github.com/DXgovernance/dxvote/pull/478#discussion_r769497493

rossneilson:
Looks like this is it, just try running the guilds and you'll see it crashes
Same happens in the deploy preview

Screenshot 2021-12-15 at 10 37 58

rossneilson:
Update for the record, the proposals show correctly when no local cache is present but crashes when there is one (not an issue before this PR although not obvious why anything here is related to cache)

rossneilson (dismissed):

mprasanjith (approved):

PR #490, opened 2021-12-14
fix: Preventing looped retry of title fetch on client

rossneilson:
- Adding more failed proposal titles
- If client then only attempt fetch once
- If title is already there from on chain data then skip

This should speed up loading of DXvote on chains like xDAI where multiple proposals with ipfs issues were slowing down loading
Also since we now have a large promise.any if it doesn't resolve it likely means it doesn't exist

### DIFF
--- a/src/cache/index.ts
+++ b/src/cache/index.ts
@@ -1835,8 +1835,7 @@ export async function getProposalTitlesFromIPFS(
       '0xfb15b6f9e3bf61099d20bb3b39375d4e2a6f7ac3c72179537ce147ed991d61b4',
     ];
 
-    // If the script is running on the client side and it already tried three times, continue.
-    console.log(proposal.title);
+    // If the script is running on the client side and it already tried once, or has the title, continue.
     if (
       invalidTitleProposals.indexOf(proposal.id) > -1 ||
       retryIntent > 3 ||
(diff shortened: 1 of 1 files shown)


src/cache/index.ts
```diff
@@ -1827,13 +1827,21 @@ export async function getProposalTitlesFromIPFS(
...
       // TODO: Fix this xdai proposal IPFS content
       '0xfb15b6f9e3bf61099d20bb3b39375d4e2a6f7ac3c72179537ce147ed991d61b4',
     ];
 
-    // If the script is running on the client side and it already tried three times, continue.
-    if (invalidTitleProposals.indexOf(proposal.id) > -1 || retryIntent > 3) {
+    // If the script is running on the client side and it already tried once, or has the title, continue.
+    if (
+      invalidTitleProposals.indexOf(proposal.id) > -1 ||
+      retryIntent > 3 ||
```
AugustoL, line 1841: The comment says "it already tried once", so it should be == 0

AugustoL (changes requested):

src/cache/index.ts
rossneilson: == 0 doesnt attempt to load it at all. Which may not be bad, it would load faster but I think one attempt makes sense to still catch daostack proposals without remote cache update

AugustoL (approved):

PR #493, opened 2021-12-14
Feature/calculate worker proposal rep from mainnet

### LINKED ISSUE #398: REP calculation using mainnet
I was under the assumption that we used xdai REP total for calculation as I thought the total supply was higher there, however this is not the case. 

We actually should be using the mainnet total supply for this calculation, meaning we need to connect to two RPCs at once. 

Since in future (#76) we want to use this feature anyways it is something we should implement in a generic way now anyways. 

REP created with current implementations will also need a slight topping up at some point.

mprasanjith:
The REP amount when creating a worker proposal is now calculated from Ethereum Mainnet, instead of currently active network.

The existing mainnet cache is used for REP calculations. If for some reason, this cache is out-of-date (more than 7 days old), the user is warned. In the future we could look at getting the updated blocks from mainnet before calculating.

The REP numbers have been cross-checked with the Worker Compensation Calculator Google sheet we were using.

Closes #398.

### DIFF
--- a/src/hooks/useMainnetRep.ts
+++ b/src/hooks/useMainnetRep.ts
@@ -1,0 +1,125 @@
+import { useWeb3React } from '@web3-react/core';
+import { useCallback, useEffect, useState } from 'react';
+import { MAINNET_WEB3_ROOT_KEY } from '../components/MainnetWeb3Manager';
+import { useContext } from '../contexts';
+import { bnum, ZERO_ADDRESS } from '../utils';
+
+const useMainnetRep = (
+  userAddress: string = ZERO_ADDRESS,
+  atBlock: number = 0,
+  atTime: number = 0
+) => {
+  const {
+    context: { configStore, ipfsService, daoStore },
+  } = useContext();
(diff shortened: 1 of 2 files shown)


rossneilson:
A while ago hamza implemented a useRep hook where we use memo, we could maybe leverage this here also or maybe even reuse that hook at some point. But fine to move this to a separate issue. Let me know
Otherwise looks great!

rossneilson (approved):

mprasanjith:
@rossneilson After checking again, I don't think we can use the current memoized `useRep` hook to calculate REP at a certain date, as it keeps the data cached by the block number. Changing the current behavior will need some significant changes to how the hook works.

rossneilson:
> @rossneilson After checking again, I don't think we can use the current memoized `useRep` hook to calculate REP at a certain date, as it keeps the data cached by the block number. Changing the current behavior will need some significant changes to how the hook works.

Sounds good, lets split it up and leave it till later

rossneilson (approved):

PR #494, opened 2021-12-14
Encode genericCall only for WalletScheme, not multicall

AugustoL:


src/pages/ProposalSubmission/Custom.tsx
```diff
@@ -305,7 +305,11 @@ const NewProposalPage = observer(() => {
           } else {
             callData = call.dataValues[0];
           }
-          if (callToController && call.to !== networkContracts.controller) {
+          if (
+            schemeToUse.type === 'WalletScheme' &&
```
rossneilson: Dont we have an isWalletScheme() function?

rossneilson (approved):

PR #503, opened 2021-12-16
Updating/removing packages & updating node

### LINKED ISSUE #505: Update node version to v16
Currently we use and require node version v14.17.6
v14 has moved in to maintenance and so we should be upgrading to v16 to use an active version with latest updates which will last us through to the end of 2022.
We should properly test all functions work fine and have proper QA time. 
Github actions will also need updated to the new version and proper heads up should be given to the dev team before merging to ensure everyone updates appropriately. 
We should also test that deterministic build hashing works as expected.
Screenshot 2021-12-16 at 19 08 43

rossneilson:
- Removed some unneeded packages
- Updated packages that have more efficient releases
- Removed unused parts of contracts files
- Updated ethers to tree shaken v5

Closes #505 As I also updated node version

Made an issue for ipfs upgrade #508


rossneilson:
Not the biggest difference but better than nothing

Before:
Screenshot 2021-12-16 at 22 58 49

After:
Screenshot 2021-12-16 at 22 00 10

rossneilson:
DONT MERGE UNTIL THOROUGHLY TESTED

- [x] QA testing by power users
- [x] Build hash testing

AugustoL (approved):

PR #507, opened 2021-12-16
Feature/improve guilds network management

### LINKED ISSUE #501: Improve network switching in guilds
Currently, guilds network switching has some strange behaviour where the network switching only really works if you connect your wallet to the default network, then switch the connected wallet.

To-do
- Add network to the URL the same as in DXvote
- Fix switching with no connected wallet

Additional related fixes 
- Change default network in guilds to rinkeby temporarily
- FIx ENS PFP loading on networks other than mainnet

mprasanjith:
This introduces a bunch of changes and bugfixes around how network switching works in guilds.

- Handle network switching on the URL
- Fixes to network switching when using injected wallets
- Fix ENS profile picture loading when not on mainnet.
- Add redirect to rinkeby when visiting guilds by default. This is temporary.
- Implement a global error boundary. (We should improve how this looks in the future.)

Fixes #501.


src/components/Guilds/ErrorBoundary/GlobalErrorBoundary.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+class GlobalErrorBoundary extends React.Component<{}, { hasError: boolean }> {
```
hhamud, line 3: Why are we using class style react component here?

src/pages/Metadata.tsx
```diff
@@ -91,7 +91,7 @@ export const CreateMetadataPage = observer(() => {
   } = useContext();
 
   const history = useHistory();
-  const { proposalType } = useParams();
+  const { proposalType } = useParams<{ proposalType?: any }>();
```
hhamud: Do we not know the ProposalTypes?, best to avoid using any type in general

src/hooks/useParams.ts
```diff
@@ -27,7 +27,7 @@ export const useParams = (searchType: string, defaultSearchtype: string) => {
     history.push({
       location: location.pathname,
       search: `${searchType}=${getParams}`,
-    });
+    } as any);
```
hhamud: I don't understand why this is needed?

src/components/Header/index.tsx
```diff
@@ -49,7 +49,7 @@ const WarningDev = styled.div`
 `;
 
 const Header = observer(() => {
-  const NavItem = withRouter(({ route, history, children }) => {
+  const NavItem = withRouter(({ route, history, children }: any) => {
```
hhamud: is it not possible to avoid using the "any" type?

src/components/common/Button.tsx
```diff
@@ -22,8 +22,8 @@ export const Button = styled.button`
   }
 `;
 
-export const LinkButton = withRouter(
-  ({ route, history, children, ...rest }) => {
+export const LinkButton: any = withRouter(
+  ({ route, history, children, ...rest }: any) => {
```
hhamud: is it not possible to avoid using the "any" type?

hhamud (changes requested):

src/components/Guilds/ErrorBoundary/GlobalErrorBoundary.tsx
mprasanjith: Weirdly, error boundaries cannot be implemented with functional components as of yet, according to the offical [React docs](https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries) so had to use a class component here.

mprasanjith:
@hhamud I added the `@types/react-router-dom` library to get types support for React Router and a bunch of typing errors popped up. I didn't wanna add types to the older code so I just used `any`. (Kinda lazy on my part.) I'll add proper types to these now.

src/hooks/useParams.ts
mprasanjith: This was throwing a type error and I didn't realize what's wrong here when I typed it as `any`. We have a bug in this code. This should be
```js
history.push({
  pathname: location.pathname,
  search: `${searchType}=${getParams}`,
});
```

It seems we're not using this whole useParams hook though, so I'll remove this whole file.

rossneilson (dismissed):

hhamud (dismissed):

rossneilson (approved):

PR #509, opened 2021-12-17
Vesting contract modal/#480

MiltonTulli:
Display vesting contract modal info. 
Add redeem action


src/pages/User.tsx
```diff
@@ -150,25 +189,43 @@ const UserPage = observer(() => {
...
+
+      {tokenVestingContracts.length ? (
+        tokenVestingContracts.map((contract, i, arr) => {
+          return (
+            <ListRow
+              clickable
+              key={contract.address}
+              borderBottom={i < arr.length - 1}
+              onClick={() => {
+                if (!contract.address) return;
```
rossneilson: Lets move this into a function outside the onclick

src/pages/User.tsx
```diff
@@ -57,6 +69,33 @@ const UserPage = observer(() => {
...
+            ...contract,
+            value: bnum('0'),
+          };
+        }
+      })
+    );
+    contracts.then(setTokenVestingContracts);
+  };
+
+  useEffect(updateUserVestingContracts, []); // eslint-disable-line
```
rossneilson: I think we should have `userAddress ` in the brackets of the useEffect so when the user goes from one profile to another it updates the contracts. Currently it does not

src/pages/User.tsx
MiltonTulli: Done! Thanks

src/pages/User.tsx
MiltonTulli: Yes, great catch! Done

src/services/DaoService.ts
```diff
@@ -357,4 +358,17 @@ export default class DaoService {
       '0'
     );
   }
+
+  redeemVestingContractDxd(contractAddress: string): PromiEvent<any> {
```
rossneilson: Not sure if there is maybe a better location for this since it's not super related to the dao 🤔

src/hooks/useContract.ts
```diff
@@ -0,0 +1,19 @@
+import { useWeb3React } from '@web3-react/core';
```
rossneilson, line 1: Could have sworn this hook existed before

rossneilson (dismissed):
Looks fantastic, just a few small things

src/services/DaoService.ts
rossneilson: Likely fine here but if you want to move it, it could likely be its own service

MiltonTulli: @rossneilson  `./src/services/TokenVestingService.ts` sounds good?

rossneilson (approved):

PR #511, opened 2021-12-17
Transaction Modal

### LINKED ISSUE #473: Transaction Modal
Currently DXvote does not provide an insight of a signed transaction. The new modals will allow the user to have more insight about the state of the transaction and allows access to etherscan:
Bildschirmfoto 2021-12-07 um 18 59 46

### Waiting for Confirmation:
Bildschirmfoto 2021-12-07 um 19 00 22
This Modal needs to appear before metamask pops up. it indicates to the user that he/she needs to sign the transaction in her/his wallet.

### Transaction Submitted:
Bildschirmfoto 2021-12-07 um 19 01 28
Comes after the user has signed/broadcasted the transaction via wallet/metamask. If a user clicks on View on Explorer he/she will get redirected (with a new tab) to etherscan or blockscout. The user can close the modal with the close icon or the close button.

### Transaction Rejected:
Bildschirmfoto 2021-12-07 um 19 03 52

Figma design specification:
https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=1239%3A4052

Check out how the modal should work with our playground by joining the guild via staking:
https://www.figma.com/proto/cjLng0AbOKuDj9r5oZVDmR/Wireframe?page-id=747%3A4460&node-id=747%3A4461&viewport=499%2C48%2C0.09&scaling=min-zoom&starting-point-node-id=747%3A4461&show-proto-sidebar=1

hhamud:
* Closes  #473 
* Refactors common Modal component
* Creates additional modal components for the different transaction states
* Creates new `Flex` and `Circle` components
* Creates hook and context for using modal anywhere throughout the guild app


src/components/Guilds/common/Modal/transaction.tsx
```diff
@@ -0,0 +1,128 @@
...
+type TransasctionModalProps = Pick<ModalProps, 'isOpen' | 'onDismiss' | 'onCancel'>;
+
+export const Container = styled.div`
+  display: flex;
+
+  flex-direction: column;
+
+  align-items: center;
+
+  padding: 24px;
```
luchux: spaces between lines? . Colors variables should be theme.

src/components/Guilds/common/Modal/transaction.tsx
```diff
@@ -0,0 +1,128 @@
...
+  </div>
+);
+
+//@TODO: make dynamic
+
+const WaitingForTransactionContent: React.ReactElement = (
+  <Container>
+    <PendingCircle height="86px" width="86px" color="black" />
+
+    <div style={{ marginBottom: '16px', marginTop: "16px" }}>
```
luchux: no inline styles pls...

src/components/Guilds/common/Modal/transaction.tsx
```diff
@@ -0,0 +1,128 @@
...
+      <ContainerText>Waiting For Confirmation</ContainerText>
+      <ContainerText lineHeight="20px" fontSize="14px" fontWeight="500">
```
luchux: why so many props here? 
is possible to extend components in styled components. So each version could be an extended version with its own css styles for height, font-size, etc. For now ok, but careful with this approach of styling by props

luchux (dismissed):

src/components/Guilds/common/Modal/transaction.tsx
hhamud: im not sure why yarn formatted it like that. I'll fix it manually at the end.

src/components/Guilds/common/Modal/transaction.tsx
hhamud: I thought it would be excessive to create a component just to set the margins. I will create a unique component for this.

src/components/Guilds/common/Modal/transaction.tsx
hhamud: You're right I got lazy with this. I will extend the components instead.

src/components/Guilds/common/Modal/transaction.tsx
luchux: yes, although is only used once, is good idea to encapsulate it in a styled component. You can extend other component with styles too if you need to add that margin but in this case is a container for more than 1 element.

rossneilson:
The changes here to modal affect other modals also, not something we should be doing.

This should likely be a higher level utility we can use to display the modal from anywhere in the app without needing to render it in the related component. We can move this work to another issue, your call, but it should be more generic

rossneilson:
CSS issues here on normal modal, first image is from preview deploy, second is develop
Screenshot 2022-01-12 at 18 39 55
Screenshot 2022-01-12 at 18 40 30

rossneilson:
Also not sure this is quite what I was meaning by my previous comments, although its an improvement by moving the logic into its own component.
What would be better is rendering this on the highest level beside the contexts and using a hook exposing a method to pass a transaction and make the modal appear with it's details. This would mean it could be called from anywhere in the code similar to how we do or will do with toast notifications.

hhamud:
> Also not sure this is quite what I was meaning by my previous comments, although its an improvement by moving the logic into its own component. What would be better is rendering this on the highest level beside the contexts and using a hook exposing a method to pass a transaction and make the modal appear with it's details. This would mean it could be called from anywhere in the code similar to how we do or will do with toast notifications.

That would be data integration and would be in the next stage. At this point the main aim is to  ensure the CSS and modal changes that this PR makes wouldn't break other components. I'll fix the CSS issues that you've mentioned and then rework the changes that you require in a different PR.

mprasanjith:
image

I suppose we can close the Lock DXD panel at this point? Feel free to ignore if this isn't a scenario that would actually happen after wiring up data.

src/GuildsApp.tsx
```diff
@@ -38,58 +39,60 @@ const GuildsApp = () => {
...
-                  to="/rinkeby/0x9cdc16b5f95229b856cba5f38095fd8e00f8edef"
-                />
-                <Redirect
-                  exact
-                  from="/:chain_name"
-                  to="/:chain_name/0x9cdc16b5f95229b856cba5f38095fd8e00f8edef"
-                />
-                <Route exact path="/:chain_name/:guild_id">
-                  <GuildsContextProvider>
+            <TransactionModalProvider>
```
rossneilson: Is there a reason this is above Global style?

rossneilson:
I'm seeing some strange behaviour in mobile view that isn't present in develop. 
I also think the modal is being rendered underneath in the next screen in mobile

Screenshot 2022-01-17 at 10 20 03

src/hooks/useFilterCriteria.ts
```diff
@@ -113,7 +113,7 @@ export const useFilterCriteria = (): useFilterCriteriaReturns => {
     ]);
 
     setLoading(false);
-  }, []);
+  }, [daoStore]);
```
rossneilson: Dont understand why this is here

src/GuildsApp.tsx
hhamud: no, I just picked a spot and slotted it in. I'll place it somewhere else

hhamud:
> I'm seeing some strange behaviour in mobile view that isn't present in develop. I also think the modal is being rendered underneath in the next screen in mobile
> 
> Screenshot 2022-01-17 at 10 20 03

This weird behaviour happens also in develop, 
image

It's because of the description proposal text not wrapping around within the box, so this causes all the modals to break.

Regarding the second issue of the modal appearing behind the lock dxd modal, madusha had pointed it out [here](https://github.com/DXgovernance/dxvote/pull/511#issuecomment-1014326531) and the same reason applies here as well, I chose to create the modal there only for reviewing purposes.

Once this PR goes through, then the next stage can go ahead and I can remove the modal from triggering at that spot and ensure that the modal only appears after closing the lock dxd modal notification.

src/hooks/useFilterCriteria.ts
hhamud: Good catch there, I cannot remember why I changed that. I'll remove it.

rossneilson (dismissed):
Looks good, nice implementation!
Could you make some follow up issues for the parts mentioned in comments?

rossneilson (approved):

PR #520, opened 2021-12-21
Feature/ens resolving optimizations

### LINKED ISSUE #492: Optimize ENS resolving
Right now resolved ENS names and avatar URLs are not cached anywhere. As this is quite expensive to resolve, we should look at caching it in session / local storage. These cached values should also expire after some time.

mprasanjith:
Resolving ENS names and avatars are expensive operations. This PR caches the fetched ENS data in local storage with expiration times of 1 day. [Fixes #492.]

This also adds two new hooks (`useLocalStorage` and `useLocalStorageWithExpiration`) that would come in handy in the future for anyone using local storage.

After trying to have a more complex data structure with a single key in local storage, I reverted to using a flat structure. It's quite hard to keep localStorage and React state in sync when there are multiple parallel reads and writes to the same key it seems. Hope this is good enough.

image

(The undefined key is intentional here. We have a scenario where the useENSAddress hook is passed `undefined` as the address.)


rossneilson (approved):
Nice implementation, very clean and works well. Likely worth stress testing some more once we have a bunch of ENS being shown everywhere but looks great.

PR #523, opened 2021-12-22
transaction counter on account button

luchux:
UI/Ux for badge into account button. 

TODO: When we connect to the web3 transactions to show in the account modal, just passing the length of the active transactions inside the badge will connect it with real data.
Screen Shot 2021-12-22 at 4 06 47 PM


hhamud (commented):
the code looks good to me but best to get geronimo's input for the design implementation of it.

hhamud (approved):

PR #526, opened 2021-12-27
Add PendingAction enum, fix redeem state for contribution reward

AugustoL:


rossneilson (approved):
Looks good, nice seeing the enum, looking forward to getting rid of old schemes and all their strange edge cases

src/components/Proposal/Status/index.tsx
```diff
@@ -27,23 +27,26 @@ const Status = () => {
...
-    daoService.execute(proposalId);
+  const executePendingAction = function (pendingAction) {
+    switch (pendingAction) {
+      case 4:
+        daoService.redeemContributionReward(
+          networkContracts.daostack.contributionRewardRedeemer,
+          scheme.address,
+          scheme.votingMachine,
+          proposalId,
+          proposal.to[0]
```
rossneilson, line 38: Think this is incorrect, Dave seems to think it should be current account being sent here, not the `to` address which is just the controller
#537

PR #527, opened 2021-12-27
fix: Fixing executed proposal order

### LINKED ISSUE #429: Ordering issue on new schemes
It seems ordering is again broken.

New scheme proposals are being sorted upside down on executed proposals. Most easily seen on xdai where there is a mix of old and new schemes. Old schemes are fine but new ones are upside down and difficult to find without filtering. 

Screenshot 2021-11-26 at 11 35 09

rossneilson:
This was fixes previously but I think was removed when migrating to a hook.
Also removing old unused code

Closes #429

### DIFF
--- a/src/hooks/useFilterCriteria.ts
+++ b/src/hooks/useFilterCriteria.ts
@@ -18,11 +18,15 @@ interface useFilterCriteriaReturns {
   loading: boolean;
 }
 
-export const orderByNewestTimeToFinish = (a: any, b: any) =>
-  a.finishTime - b.finishTime;
+export const orderByNewestTimeToFinish = (
+  a: ProposalsExtended,
+  b: ProposalsExtended
+) => a.finishTime - b.finishTime;
 
-export const orderByOldestTimeToFinish = (a: any, b: any) =>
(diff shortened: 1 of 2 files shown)


src/hooks/useFilterCriteria.ts
```diff
@@ -19,6 +18,12 @@ interface useFilterCriteriaReturns {
   loading: boolean;
 }
 
+export const orderByNewestTimeToFinish = (a: any, b: any) =>
```
MiltonTulli: Not sure if the intention is not to use a lot of types. But `(a: ProposalsExtended, b: ProposalsExtended)` sounds appropriate. ( for both `orderByNewestTimeToFinish` and `orderByOldestTimeToFinish` )

MiltonTulli (dismissed):

src/hooks/useFilterCriteria.ts
rossneilson: Nice spot, I was just copy pasting. Fixed now

MiltonTulli (approved):

PR #534, opened 2021-12-29
Feature/create pagination proposals hook

### LINKED ISSUE #528: Add pagination to proposal fetching
Currently, we are fetching proposal ids then just fetching every proposal object for all ids. This is fine now but of course wont scale well. (src/hooks/guilds/useProposals.ts)
We don't yet have designs for pagination in guilds yet but we can implement it on the backend for now. 

- Get all proposalIds for the guild
- Fetch further proposal data based on current page passed to hook
- Save already fetched proposal objects

We should calculate the number of pages based on total proposals and pagesToDisplay. The hook could then return the proposal objects in an array based off of `currentPage`. This could be in the form of a usePaginatedProposals() which uses other hooks which are responsible for storing/caching the total proposals available.

MiltonTulli:
## Summary
Create usePaginatedProposals hook

## Issue(s)
https://github.com/DXgovernance/dxvote/issues/528


rossneilson (approved):
Looks fantastic, a few things spring to mind about how we can improve this but need to consider more about how we are updating and showing data in the future. For now a great start using contexts and splitting up ids from proposal data

mprasanjith (approved):

PR #550, opened 2022-01-06
Feature/multichain provider

mprasanjith:
This PR lets DXvote to connect to all networks at the same time.

We previously had code to have a persistent Mainnet connection for use cases such as accessing ENS names, NFT avatars and mainnet REP numbers. This expands on the previous implementation to allow DXvote to connect to all networks at the same time. To do this, existing web3-react based implementation has been replaced by Ethers.js. Existing code in both DXvote and Guilds has been updated to reflect the changes.

The new `useJsonRpcProvider` hook allows to obtain a Ethers JsonRpcProvider to any network just by passing in the chain ID.
```js
const provider = useJsonRpcProvider(chainId)
```

If no parameter is passed, it returns the current wallet Web3 connection. 
```js
const walletWeb3Provider = useJsonRpcProvider()
```


luchux (approved):
works nice! the cool looks good, improvements are coming into future PRs!

PR #557, opened 2022-01-08
Get guild instance data - Create a single contract instance on context.

### LINKED ISSUE #539: Get guild instance data
In guilds we currently collect proposal data but we also need to do an initial load of guild data on load 

- getToken()
- getPermissionRegistry()
- getName()
- getProposalTime()
- getTimeForExecution()
- getMaxActiveProposals()

Likely best that we store all this in a context for basic guild instance data

MiltonTulli:
### Summary: 
- Create Guild Config context
- Store contract instance
- Pull guild contract data and store it inside context. 
- Refactor `useProposals` and `usePaginatedProposals` hooks to use context contract instance.

### Issue(s): 
https://github.com/DXgovernance/dxvote/issues/539


mprasanjith (changes requested):
I'd suggest to move the ConfigContext and Provider to `src/contexts/Guilds` and `useGuildConfig` hook to `src/hooks/Guilds` for consistency.

MiltonTulli:
> I'd suggest to move the ConfigContext and Provider to `src/contexts/Guilds` and `useGuildConfig` hook to `src/hooks/Guilds` for consistency.

Done. I put `useGuildConfig` under `hooks/Guilds`

rossneilson (dismissed):

luchux (dismissed):
awesome work!

src/contexts/Guilds/config.tsx
```diff
@@ -0,0 +1,148 @@
...
+        token: config.token,
+        permissionRegistry: config.permissionRegistry,
+        name: config.name,
+        proposalTime: config.proposalTime,
+        timeForExecution: config.timeForExecution,
+        maxActiveProposals: config.maxActiveProposals,
+      });
+    };
+    initialize();
+  }, [web3Context.chainId, address]); // eslint-disable-line react-hooks/exhaustive-deps
```
mprasanjith: Is it possible to fix this without disabling the eslint rule?

mprasanjith (dismissed):
Looks good! I think this PR also contains changes that we should update to use SWR. Pinging @luchux for visiblity.

rossneilson (dismissed):
Looks great, lets follow it up with SWR after also

luchux:
@mprasanjith @MiltonTulli 
We are changing a bunch of the way we read the data in Guilds page and each Proposal page. 
Do you guys think we should wait a bit with this PR, to see when we need the context? useEtherSWR removes the need of keeping a context (if we use directly the data as is stored in the EtherSWR cache,  and also if we add a middleware to EtherSWR we can manipulate and store the data in a post-processed after fetch way - like adding metadata -, then we wont need a context). But still would be interesting to analyze if we need stuff like that or not. 

I would not discard this changes, as if we remove useEtherSWR this is the way to go, but will put on hold (and not even try to fix the conflicts). 

What are your thoughts?

hhamud:
> @mprasanjith @MiltonTulli We are changing a bunch of the way we read the data in Guilds page and each Proposal page. Do you guys think we should wait a bit with this PR, to see when we need the context? useEtherSWR removes the need of keeping a context (if we use directly the data as is stored in the EtherSWR cache, and also if we add a middleware to EtherSWR we can manipulate and store the data in a post-processed after fetch way - like adding metadata -, then we wont need a context). But still would be interesting to analyze if we need stuff like that or not.
> 
> I would not discard this changes, as if we remove useEtherSWR this is the way to go, but will put on hold (and not even try to fix the conflicts).
> 
> What are your thoughts?


I say we place this into draft mode for now and work on extending EtherSWR

MiltonTulli:
> > @mprasanjith @MiltonTulli We are changing a bunch of the way we read the data in Guilds page and each Proposal page. Do you guys think we should wait a bit with this PR, to see when we need the context? useEtherSWR removes the need of keeping a context (if we use directly the data as is stored in the EtherSWR cache, and also if we add a middleware to EtherSWR we can manipulate and store the data in a post-processed after fetch way - like adding metadata -, then we wont need a context). But still would be interesting to analyze if we need stuff like that or not.
> > I would not discard this changes, as if we remove useEtherSWR this is the way to go, but will put on hold (and not even try to fix the conflicts).
> > What are your thoughts?
> 
> I say we place this into draft mode for now and work on extending EtherSWR

Sure. I prefer closing this PR since we are going to implement SWR lib and that won't require any config context or contract methods call as I have done here. That was something that wasn't clear when started working on this. Will close this pr and work on a new implementation. Thanks!

PR #560, opened 2022-01-10
Fix/network switch issues

AugustoL:


rossneilson (dismissed):

src/stores/DaoStore.ts
```diff
@@ -705,44 +704,38 @@ export default class DaoStore {
       redeemsDaoBounty: [],
     };
 
-    for (const votingMachineName in votingMachines) {
-      const votingMachine = votingMachines[votingMachineName];
+    Object.keys(cache.votingMachines).map(votingMachine => {
+      console.log(votingMachine);
```
mprasanjith: Maybe we can get rid of this console.log?

rossneilson (approved):

PR #565, opened 2022-01-12
Feature/ether swr implementation

luchux:
Does:
- merges missing work into develop to use data with etherSWR 
- refactors some work on ProposalCard


src/components/Guilds/ProposalCard/index.tsx
```diff
@@ -91,11 +96,25 @@ const ProposalStatusWrapper = styled.div`
 `;
 
 interface ProposalCardProps {
-  proposal: Proposal;
+  id: any;
```
mprasanjith, line 99: Can we make this `string`?

mprasanjith (changes requested):

mprasanjith (approved):

rossneilson:
Is this polling at all? Like does it poll by default?

PR #567, opened 2022-01-12
Feature/guilds swr global config

luchux:
Does: 
- sets a global config wrapping the APP with a EtherSWR component that passes config stuff. 
We could extend this as per routes, or different parts of the app. So far, so good.

### DIFF
--- a/src/GuildsApp.tsx
+++ b/src/GuildsApp.tsx
@@ -5,7 +5,9 @@ import {
   useHistory,
   Redirect,
 } from 'react-router-dom';
+//import { useParams } from 'react-router-dom';
 import { ThemeProvider } from 'styled-components';
+import { EthSWRConfig } from 'ether-swr';
 import { Container } from './components/Guilds/common/Layout';
 
 import Header from './components/Guilds/Header';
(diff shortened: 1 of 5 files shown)


src/GuildsApp.tsx
```diff
@@ -31,30 +39,46 @@ const GuildsApp = () => {
...
-                </Route>
-              </Switch>
-            </Container>
+            <EthSWRConfig
+              value={{
+                web3Provider: provider,
+                ABIs: new Map([
+                  [
+                    // If we need to generate this for each guildId we may need this wrapping after each route
+                    '0x9cdc16b5f95229b856cba5f38095fd8e00f8edef',
```
rossneilson, line 48: So we are definitely going to need this to be dynamic. We are also going to need the abi to be potentially different in future.
We will have some different variances and implementations ontop of the base erc20 contract.

src/GuildsApp.tsx
```diff
@@ -31,30 +39,46 @@ const GuildsApp = () => {
...
+                    ERC20GuildContract.abi,
+                  ],
+                ]),
+                // If we need different polls in each part of the app we can reconfigure this as param when fetching
+                // or a different wrapp for each. Remove this line when we are all set up.
+                refreshInterval: 30000,
```
rossneilson, line 54: I think better to have this be 0 as default

src/GuildsApp.tsx
rossneilson: So when we useSWR we define which ABI is used by the guild Id right?

PR #568, opened 2022-01-12
[WIP] Feature/guild data with swr

MiltonTulli:
Use SWR to get guild data


luchux:
@MiltonTulli Lovely! 
i have a new PR where im doing config at each route level, but not need really. So if mine get merged first, then we can fix conflicts in favour of your approach. That useGuildSWR Config looks rad ;)

PR #569, opened 2022-01-13
Feature/guilds swr proposal page

luchux:
Doing: 
- moving Proposal page to SWREther
- moving global config out of each SWREther, and instead wrapping each route (so we can pass whatever we need)
- trying different interval polls.


src/GuildsApp.tsx
```diff
@@ -47,11 +54,39 @@ const GuildsApp = () => {
                 />
                 <Route exact path="/:chain_name/:guild_id">
                   <GuildsContextProvider>
-                    <GuildsPage />
+                    <EtherSWRConfig
+                      value={{
+                        web3Provider: provider,
+                        ABIs: new Map([
+                          [
```
rossneilson, line 61: We'll need some way of making this map of address and ABI dynamic

rossneilson (approved):

PR #573, opened 2022-01-16
Quiet Ending Period Sorting Incorrectly

### LINKED ISSUE #572: Quiet Ending Period Sorting Incorrectly
**Description**
The Atlantic World Proposal needed to be voted down in the last day because of an incorrect address. This means the proposal went into the "Quiet Ending Period" because a proposal needs to stay either FOR or AGAINST for 2 days. This changes the state on genesis protocol and on DXvote its labeled as "Quient Ending Period"

Currently, proposals in this state show up below the Expired Proposals, when in reality they should be near the top, or in the chronological order of when the proposal will be finished.

Screen Shot 2022-01-14 at 3 49 54 PM

At the minimum, it should be moved up ahead of all other states except for boosted (In queue, pre-boosted, pending boost)

hhamud:
* Closes #572 
* Removed `QuiteEndingPeriod` from `StateEarliestUnder10` and `StateEarliestAbove10` and created its own filtered array.

The filtering was previously amended by @luchux in discussion with the DXvote power users however I can't remember why exactly the criteria was chosen, so would need @luchux guidance on this one.

### DIFF
--- a/src/hooks/useFilterCriteria.ts
+++ b/src/hooks/useFilterCriteria.ts
@@ -38,6 +38,7 @@ export const useFilterCriteria = (): useFilterCriteriaReturns => {
 
   useEffect(() => {
     const allProposals = daoStore.getAllProposals();
+
     // (QuitedEndingPeriod || Queded) && positiveVotes >= 10% (Ordered from time to finish, from lower to higher)
     const stateEarliestAbove10 = allProposals
       .filter(proposal => {
@@ -46,10 +47,7 @@ export const useFilterCriteria = (): useFilterCriteriaReturns => {
         ).totalSupply;
 
         return (
-          (proposal.stateInVotingMachine ===
-            VotingMachineProposalState.QuietEndingPeriod ||
-            proposal.stateInVotingMachine ===
-              VotingMachineProposalState.Queued) &&
+          proposal.stateInVotingMachine === VotingMachineProposalState.Queued &&
           proposal.positiveVotes
             .div(repAtCreation)
             .times(100)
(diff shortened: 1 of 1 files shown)


src/hooks/useFilterCriteria.ts
```diff
@@ -46,10 +47,7 @@ export const useFilterCriteria = (): useFilterCriteriaReturns => {
         ).totalSupply;
 
         return (
-          (proposal.stateInVotingMachine ===
-            VotingMachineProposalState.QuietEndingPeriod ||
-            proposal.stateInVotingMachine ===
-              VotingMachineProposalState.Queued) &&
+          proposal.stateInVotingMachine === VotingMachineProposalState.Queued &&
```
luchux, line 50: @hhamud  The criteria was coming from a conversation with @AugustoL and @rossneilson on the issue when adding these filters. 
Was the initial criteria. If the issue is for Quiet Ending Period sorting wrong, this wouldn't be the problem. 
I am not sure of the origin of this issue, but lets keep a conversation with the guys to be sure we don't break the initial set up for the filtering of the first render.

src/hooks/useFilterCriteria.ts
```diff
@@ -38,6 +38,7 @@ export const useFilterCriteria = (): useFilterCriteriaReturns => {
 
   useEffect(() => {
     const allProposals = daoStore.getAllProposals();
+
     // (QuitedEndingPeriod || Queded) && positiveVotes >= 10% (Ordered from time to finish, from lower to higher)
```
luchux: @hhamud if you end up removing that criteria in the filter, change this comment. 
cheers!

rossneilson (approved):

PR #575, opened 2022-01-17
Feature/proposal page sidebar data wireup

mprasanjith:
This PR wires up the proposal page sidebar information box to actual data.

In addition:
- Some unused sidebar components and hooks have been removed.
- All SWR hooks have been moved to its own subfolder.
- Proposal date parsing has been moved into a SWR middleware.
- A middleware has been introduced to implement logging SWR calls. Enable verbose logging on browser devtools to see these logs :)


src/components/Guilds/ProposalPage/ProposalInfoCard/index.tsx
```diff
@@ -1,5 +1,9 @@
+import moment, { unix } from 'moment';
+import React, { useMemo } from 'react';
 import { FiCheck, FiInbox } from 'react-icons/fi';
+import Skeleton from 'react-loading-skeleton';
 import styled from 'styled-components';
+import { useProposal } from '../../../../hooks/Guilds/proposals/useProposal';
```
luchux: @mprasanjith This hook is not the one using EtherSWR. 
that one is located at hooks/Guilds/useProposal
I just let it there because we may not be ready to move away from Guilds/proposals/useProposal

src/components/Guilds/ProposalPage/ProposalInfoCard/index.tsx
```diff
@@ -34,7 +38,40 @@ const UserInfoDetail = styled(Box)`
...
+  guildId: string;
+}
+
+const ProposalInfoCard: React.FC<ProposalInfoCardProps> = ({
+  guildId,
+  proposalId,
+}) => {
+  const { proposal, error } = useProposal(guildId, proposalId);
+
+  const startTime = useMemo(() => {
```
luchux: @mprasanjith This stuff is very similar to ProposalStatus calculation. 
Seems to me we can add this memo as part of a middleware in useEtherSWR, so that every result data is postprocessed and the attribute calculated and added, OR, we can put this as a helper in useProposal hook, or as an attribute for every proposal as part of its prototype.

luchux (dismissed):

luchux (approved):

PR #577, opened 2022-01-17
header sticky

luchux:
Does: 
- sticky navbar. 
- sticky menu on SidebarCard component for mobile. 

DRAFT: 
- on mobile seems to fail with the simple CSS position: sticky attribute. 
debugging


~
```diff
@@ -0,0 +1,480 @@
+import { BigNumber, ethers } from 'ethers';
```
rossneilson: ?

luchux: what is this? haha. let me see. wait, what?

src/components/Guilds/Header/index.tsx
```diff
@@ -33,6 +41,7 @@ const Header = () => {
 
   return (
     <HeaderWrapper as="header">
+      <div></div>
```
mprasanjith: Curious if this is something we need?

mprasanjith (changes requested):

src/components/Guilds/Header/index.tsx
luchux: yeah, part of the workaround

rossneilson (dismissed):
Please create a follow up issue for the mobile implementation

mprasanjith (approved):

PR #584, opened 2022-01-17
Add Cache page (build/verify cache in browser)

AugustoL:
Moved cache script to a cache service, to be used by the blockchain store and the cache page, the
cache page receive the options to run the cache build script in the dapp and verifies the hash of
the config received with the target hash. It also provides the .json files of the configs and cache
to be downloaded.


rossneilson:
Small note, would be awesome if we could change the title of the tab when running the cache to see that its different from DXvote to avoid accidentally closing it and also to show progress or when its finished

AugustoL:
@rossneilson I updated the configs in the source code with the one we got building it together. Now we can merge

rossneilson (approved):

PR #587, opened 2022-01-18
Get guild instance data 

### LINKED ISSUE #539: Get guild instance data

MiltonTulli:
### Summary
Get contract instance data(token address, name, proposal duration, proposal execution ) and show that on sidebar component

Screen Shot 2022-01-18 at 15 51 56


src/components/Guilds/Sidebar/StakeTokens.tsx
```diff
@@ -87,14 +93,46 @@ const ButtonLock = styled(Button)`
...
+    error,
+  } = useConfig();
+
   return (
     <GuestContainer>
       <DaoBrand>
         <DaoIcon src={dxIcon} alt={'DXdao Logo'} />
-        <DaoTitle>DXdao</DaoTitle>
+        <Loading text loading={!name && !error}>
+          <DaoTitle>{error ? 'DxDao' : name}</DaoTitle>
```
rossneilson: Dont think we need a backup placeolder when it fails here, eventually we'll have a proper error screen also

src/components/Guilds/Sidebar/StakeTokens.tsx
```diff
@@ -87,14 +93,46 @@ const ButtonLock = styled(Button)`
...
+        </Loading>
       </DaoBrand>
       <InfoItem>40% Quorum</InfoItem>
-      <InfoItem>6 days proposal duration</InfoItem>
+      <Loading text loading={!token && !error}>
+        <InfoItem>
+          {!error && (
+            <>
+              Token: {isAddress(token) ? shortenAddress(token) : ''}
```
rossneilson: We should resolve the token name at some point also but for now this is good

src/hooks/Guilds/useConfig.ts
```diff
@@ -0,0 +1,45 @@
+import useEtherSWR from 'ether-swr';
```
rossneilson, line 1: This is good, we also discussed using a wrapping style like this for all swr calls so that we can test it easier, once we have that we can move this in there also

luchux: this is ok @rossneilson, and enough for mocking ! 
because we can mock the useConfig hook.

luchux (dismissed):

src/components/Guilds/Sidebar/StakeTokens.tsx
MiltonTulli: Sure, will remove this default string.

src/components/Guilds/Sidebar/StakeTokens.tsx
MiltonTulli: I just added it because `shortenAddress` throws an error if isn't a valid address.

luchux (approved):

PR #588, opened 2022-01-18
testing setup

luchux:
Does: 
- sets up jest testing with the config as in react-app-rewired docs, to use the app package webpack. 
- creates a silly test for button to run. 
- adds commands to run > yarn tes

TODO: 
We had webpack as a dependecy (not needed separated in react-app-rewired as mentioned in the docs) but seems it has been addded for typescript and plugins such as fork-ts-checker-webpack-plugin
The problem is that this keeps two webpacks in conflict (the one of rewired and the manual added). 
I adjusted manually the webpack version to be compatible with both (rewired jest, and  fork-ts-checker-webpack-plugin). 
Would need to chat with whoever configed TS. 

Screen Shot 2022-01-18 at 6 20 10 PM
t


src/utils/tests.js
```diff
@@ -0,0 +1,6 @@
+import { ThemeProvider } from 'styled-components';
+import theme from '../theme/light.json';
+
+export const themedComponent = children => (
+  <ThemeProvider theme={theme}> {children} </ThemeProvider>
+);
```
MiltonTulli: Maybe something like`renderWithTheme` will save us some extra lines in the future. 
```
import { render } from '@testing-library/react';

export const renderWithTheme = children => {
  return render(<ThemeProvider theme={theme}> {children} </ThemeProvider>);
};

File: src/components/Guilds/common/Button/Button.test.jsx [hunk 2]
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { render, fireEvent, screen } from '@testing-library/react';
+import { Button } from './Button';
+import { themedComponent } from '../../../../utils/tests';
+
+test('renders button', async () => {
+  render(themedComponent(<Button>hola</Button>));
+  fireEvent.click(screen.getByText('hola'));
+});
```
MiltonTulli: I don't see why we should lose the `.tsx` extension on this file.

src/utils/tests.js
```diff
@@ -0,0 +1,6 @@
+import { ThemeProvider } from 'styled-components';
```
MiltonTulli, line 1: Same here. `utils.tsx`

src/components/Guilds/common/Button/Button.test.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
MiltonTulli: I think this import isn't required.

package.json
```diff
@@ -17,7 +17,8 @@
     "cy:run": "cypress run",
     "cy:run:development": "cypress run --env configFile=development",
     "cy:run:production": "cypress run --env configFile=production",
-    "test:e2e": "start-server-and-test 'yarn start --no-browser' 3000 cy:run:development"
+    "test:e2e": "start-server-and-test 'yarn start --no-browser' 3000 cy:run:development",
+    "test": "react-app-rewired test"
```
MiltonTulli: Just a very opinionated thing. If this starts to grow it will be nice to have a coverage report. Maybe not important now but something I like to have. We could add some basic flags like this:

```suggestion
    "test": "react-app-rewired test --watchAll=false --coverage --verbose"
```

MiltonTulli: Quick question. If we know that `react-app-rewired` is going down why don't we use `jest` directly?

luchux: because the config of webpack and babel that rewired does. we can chat more about it.

src/utils/tests.js
luchux: thanks!~

src/components/Guilds/common/Button/Button.test.jsx
luchux: sure

src/utils/tests.js
luchux: yeah, this one I think i need some other config cause typescript fails at least in visual code. 
does it fails if you change it?

MiltonTulli: I tested on your branch and it's working for me (also vscode). Did you try changing it for ts or tsx?

PR #591, opened 2022-01-19
Feature/guilds store user transactions

### LINKED ISSUE #582: Add Transaction Display function to Modals
image

I suppose we can close the Lock DXD panel at this point? Feel free to ignore if this isn't a scenario that would actually happen after wiring up data.

_Originally posted by @mprasanjith in https://github.com/DXgovernance/dxvote/issues/511#issuecomment-1014326531_

mprasanjith:
This PR introduces a global transactions context to keep track of all Web3 transactions invoked by the app. Fixes #591.

### Summary of changes

- A global `TransactionContext` has been added to keep track of invoked transactions.
- Transaction statuses for ongoing transactions are updated automatically in real-time.
- `TransactionModal` components has been refactored to support the new context.
- `TransactionModal` components are triggered automatically whenever needed.
- A `ToastNotificationContainer` has been added, with the guilds styles applied to push notifications.
- Transaction status toast notifications are automatically triggered and updated.
- `AccountModal` has been updated to reflect the latest UI designs in Figma.
- `AccountModal` shows the latest 5 transactions from the `TransactionContext`.

To test these changes, a temporary dummy transaction has been added to the Lock DXD action. We should refactor it to call the guilds contracts soon.

### Screenshots

Toasts: 
image
image

Account modal:
image

### Dev Notes
A new helper method has been implemented that handles all user prompts and notifications when executing a transaction.
```js
createTransaction(`Approve token spending limit`, async () => {
  return contract.approve(spenderAddress, BigNumber.from(amount))
});
```
The helper takes in a summary text of the transaction and an async function that must return an ethers.js `TransactionResponse`. The summary text is only used for descriptive purposes in the UI. The helper will prompt the user to execute the async function, and handle displaying execution progress, as well as any exceptions.


luchux (approved):

luchux:
Excellent work so far! 
I thing we should merge and then refactor the harcoded things later on, so we can keep testing it. 
BTW, worth opening a separate Issue, but we should write good unittest  for this.

PR #596, opened 2022-01-20
Feature/guild editor

rossneilson:
New headless editor for proposal descriptions
Saving json, HTML and markdown locally and we can decide how best to render it later

Closes #571

PR #597, opened 2022-01-20
New proposal type selection page

### LINKED ISSUE #563: New proposal type selection page
This is the first stage of creating a proposal where the user selects a proposal type 

The options should be defined by a config file with name, description and an array of actions.
The description is shown below as a preview. 

Check for mobile view, at time of creation there was not one made yet

https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=1926%3A5575

hhamud:
* Close #563 
* Moved container text into its own file in Typegraphy
* Created proposal type selection page
* Created generic SVG wrapper
* Created configuration file for configuring the data of the proposal types


src/components/Guilds/Filter/index.tsx
```diff
@@ -24,15 +26,41 @@ const FilterRow = styled.div`
...
+
 export const Filter = () => {
+  //@TODO talk to whoever made this how best to refactor
   const [viewFilter, setViewFilter] = useState(false);
+  const [createProposal, setCreateProposal] = useState(true);
   const { totalFilters } = useFilter();
+
+  const history = useHistory();
+  const location = useLocation();
+  console.log(location);
```
rossneilson: Console log

rossneilson:
If you take a look at the issue description we need this to be configurable so it can be easily changed and the config easily added to so we can predefine actions of proposal types in future

src/components/Guilds/Filter/index.tsx
```diff
@@ -24,15 +26,41 @@ const FilterRow = styled.div`
...
-        {isMobile && (
+        {createProposal && (
+          <ButtonContainer>
+            <Button onClick={() => setCreateProposal(false)}>
+              Proposal state
+            </Button>
+            <Button
+              variant="secondary"
+              onClick={() =>
+                history.push(location.pathname + '/proposalselection')
```
rossneilson: Could we rename this to `proposalType`
Sorry, should have included this in issue description

rossneilson (dismissed):
If you take a look at the issue description we need this to be configurable so it can be easily changed and the config easily added to so we can predefine actions of proposal types in future

rossneilson:
If my PR #596 is merged before this you could also link to the next page along which I added (/createProposal/:type) with the type on the end

scripts/build.sh
```diff
@@ -2,4 +2,4 @@
 export REACT_APP_GIT_SHA=$(echo $(git rev-parse  HEAD) | cut -c1-9)
 export REACT_APP_VERSION=$npm_package_version
 export SKIP_PREFLIGHT_CHECK=true
-yarn react-app-rewired build
+yarn react-app-rewired --max_old_space_size=4096 build
```
rossneilson, line 5: ?

src/GuildsApp.tsx
```diff
@@ -74,6 +78,25 @@ const GuildsApp = () => {
                       </EtherSWRConfig>
                     </GuildsContextProvider>
                   </Route>
+                  <Route exact path="/:chain_name/:guild_id/ProposalType">
+                    <GuildsContextProvider>
```
MiltonTulli, line 82: I know this is not specific from your pr, but shouldn't we add one single GuildsContextProvider (maybe after transactionProvider or something like that) instead of adding a new contextProvider per route?

scripts/build.sh
hhamud: I was getting JS heap memory errors during its CI build, this was a solution to the problem.

src/components/Guilds/Filter/index.tsx
```diff
@@ -24,15 +26,37 @@ const FilterRow = styled.div`
...
-        {isDesktop && <FilterMenu />}
-        {isMobile && (
+        {createProposal && (
+          <ButtonContainer>
+            <Button onClick={() => setCreateProposal(false)}>
+              Proposal state
+            </Button>
+            <Button
+              variant="secondary"
+              onClick={() => history.push(location.pathname + '/ProposalType')}
```
MiltonTulli: Maybe we can add`proposalType` to be consistent with the rest of the routes on the app.
```suggestion
              onClick={() => history.push(location.pathname + '/proposalType')}

```

src/components/Guilds/Filter/index.tsx
```diff
@@ -24,15 +26,37 @@ const FilterRow = styled.div`
...
+
   return (
     <FilterContainer>
       <FilterRow>
         <InputText placeholder="Proposal title" />
-        {isDesktop && <FilterMenu />}
-        {isMobile && (
+        {createProposal && (
+          <ButtonContainer>
+            <Button onClick={() => setCreateProposal(false)}>
```
rossneilson, line 47: Im confused, what is this button and `createProposal` value doing?

src/components/Guilds/Header/index.tsx
```diff
@@ -5,15 +5,15 @@ import { Heading } from '../common/Typography';
 import NetworkButton from './NetworkButton';
 import WalletButton from './WalletButton';
 
+//@NOTE: any reason why this is set to this value?? z-index: 999999;
```
rossneilson: Remove note

src/components/Guilds/common/SVG/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+const StyledIconWrapper = styled.div<StyledIconProps>`
+  margin: ${({ margin }) => margin || '0 8px'};
+  & svg {
+    & path {
+      fill: currentColor;
+    }
+  }
+`;
+
+const StyledIcon = React.memo((props: StyledIconProps) => {
```
MiltonTulli, line 18: Quick question! Why did you consider memorization here? I don't see any heavy stuff to memorize. Just small jsx.

src/components/Guilds/ProposalTypes/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+interface ProposalTypesProps {
+  data: ProposalTypeDescriptionProps[];
+}
+const ProposalTypes: React.FC<ProposalTypesProps> = ({ data }) => {
+  const history = useHistory();
+  const location = useLocation();
+
+  const continueUrl = location.pathname.replace(
+    '/ProposalType',
+    '/create/proposal_type'
```
rossneilson, line 154: Could we make this add the actual proposal type to the URL?

src/components/Guilds/Filter/index.tsx
hhamud: the design's changed but I did not want to remove the previous dropdown menu's completely so this would be a stop gap.

src/GuildsApp.tsx
hhamud: Yeah I thought about as well but that would be outside the scope of this PR, so I didn't want to change anything. I'll make an issue about it.

src/components/Guilds/ProposalTypes/index.tsx
hhamud: I was not sure whether or not each proposal type would have a different URL and template? or if it would all point to the same URL with predefined components?.

I'd rather leave this change outside of this scope.

src/components/Guilds/common/SVG/index.tsx
hhamud: Because it is a generic component it would be designed to be used anywhere an SVG icon is used therefore I would like to avoid re-renders whenever possible.

src/components/Guilds/Filter/index.tsx
hhamud: Good catch. I will change this

rossneilson (approved):

PR #600, opened 2022-01-21
Feature/guilds unittest proposal status

luchux:
Does: 
- refacotrs proposalStatus to use a proposalId as prop or in params get by useParams(), 
- fetch the data for the proposal from useEtherSWR (that is cached/store shape). Access proposal and renders status. 
- tests for proposalStatus with an old proposal mocked.


.github/workflows/jest-tests.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install Dependencies
+        run: yarn
+      - name: Run Jests Test
+        run: yarn test
```
MiltonTulli, line 25: We can include coverage report too
```

      - name: Generate coverage report
        uses: ArtiomTr/jest-coverage-report-action@v2.0-rc.6
        with:
            github-token: ${{ secrets.GITHUB_TOKEN }}
            # threshold: 20 # TODO: Include this when we reach aceptable coverage %

Comment (luchux) [reply .github/workflows/jest-tests.yml:25]: CRACK!

File: src/components/Guilds/ProposalCard/ProposalCard.test.tsx [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+
+jest.mock('hooks/Guilds/useProposal', () => ({
+  useProposal: () => ({
+    data: {
+      title: 'Proposal Title',
+      description: 'Proposal Description',
+      contentHash: '0x0',
+      endTime: { toNumber: () => 3 },
+    },
+    loading: false,
```
MiltonTulli: I see you mocked all Swr calls with loading key. I think it should be `isValidating` instead of `loading`. 
```
export interface SWRResponse<Data = any, Error = any> {
    data?: Data;
    error?: Error;
    mutate: KeyedMutator<Data>;
    isValidating: boolean;
}
```

MiltonTulli (approved):

PR #609, opened 2022-01-24
v1.0.9m merge to master

rossneilson:


MiltonTulli (approved):
I think we need to modify github actions condition to avoid running 2 times the same action

rossneilson:
> I think we need to modify github actions condition to avoid running 2 times the same action

Good point

PR #614, opened 2022-01-25
Add missing logic to detect stake bounty and minor bug fixes

AugustoL:
Detect and calculate remaining stake bounty, fix proposer reward to be added only on proposals that
ended with YES vote, add missing edge case in decodeProposalStatus if proposal ended with NO and
executed


skyminelabs:
Thanks @AugustoL The Redeems for Awards is now showing on my Profile page and I can Redeem. Thanks!

luchux (approved):

PR #617, opened 2022-01-26
Bugfix/guilds proposal card loading state fixes

### LINKED ISSUE #615: Fixed size proposal cards
We have a lot of movement as things load in guilds right now.
We can fix this by setting the height of placeholder loading animations to be the same as the real content

mprasanjith:
This is a few improvements to the proposal cards loading state.

- Implements the skeleton loading states inline, leading to less [layout shift](https://web.dev/cls/).
- Uses the Intersection Observer API to load the cards on-scroll as they enter the viewport, to support lazy infinite loading.
- The cards have also been updated to correctly show the Eth address and resolve the ENS name / avatar if available.
- Unit tests have been added / updated.

Fixes #615.


luchux (dismissed):

src/pages/Guilds/Guilds.tsx
```diff
@@ -55,18 +53,27 @@ const GuildsPage: React.FC = () => {
...
+                  )}
+                </InView>
               ))
+            ) : (
+              <>
+                <ProposalCard />
+                <ProposalCard />
+                <ProposalCard />
+                <ProposalCard />
+                <ProposalCard />
```
hhamud, line 75: Is there a better way to have a default loading state for when proposalIds are loading?. I'm not entirely sure what best practises here would be.

hhamud (approved):
LGTM. Just a single comment

src/pages/Guilds/Guilds.tsx
mprasanjith: I'm not sure if there's a significantly better way, as we somehow need to render an array of cards here...

PR #620, opened 2022-01-27
Feature/ether swr multi chain support

### LINKED ISSUE #618: Ether-SWR not handling network switching
Ether-SWR doesn't clear the cache when the wallet network is switched, leading to stale data and generally unpredictable behavior. We should force clear the cache when the network is switched.

- Maybe its possible to come up with a solution to still keep the previous network cache data as well, without clearing everything out?
- Ideally, this should be part of the ether-swr library. Maybe we can make an upstream PR to their repo once we have a proper solution.

mprasanjith:
This PR implements an important change to how we use Ether-SWR. The current [Ether-SWR](https://github.com/aboutlo/ether-swr/) implementation does not support network switching. This implements a workaround for this issue, by implementing a custom hook that manually mutates swr upon network switching.

**This change requires all developers to use the custom useEtherSWR hook in `src/hooks/Guilds/ether-swr/useEtherSWR.ts` instead of the one exported from the `ether-swr` library.**

Fixes #618.


src/hooks/Guilds/ether-swr/useEtherSWR.ts
```diff
@@ -0,0 +1,41 @@
...
+  key: ethKeyInterface | ethKeysInterface | etherKeyFuncInterface
+): SWRResponse<Data, Error>;
+function useEtherSWR<Data = any, Error = any>(
+  key: ethKeyInterface | ethKeysInterface | etherKeyFuncInterface,
+  config?: EthSWRConfigInterface<Data, Error>
+): SWRResponse<Data, Error>;
+function useEtherSWR<Data = any, Error = any>(
+  key: ethKeyInterface | ethKeysInterface | etherKeyFuncInterface,
+  fetcher?: any,
+  config?: EthSWRConfigInterface<Data, Error>
```
hhamud: can you fix all the type any's in another issue if you can't do it in this one

hhamud (dismissed):

src/hooks/Guilds/ether-swr/useEtherSWR.ts
mprasanjith: I copied these TS definitions over from Ether-SWR [here](https://github.com/aboutlo/ether-swr/blob/master/src/useEtherSWR.ts#L35) and they have them set to `any` with the actual types commented as well. Wasn't sure what the reason was, so decided to leave them as is, but will fix them.

mprasanjith: BTW `Data` and `Error` is intentionally `any` by default, because the users are supposed to override them to whatever type they should actually be. Both SWR and Ether-swr uses `any` as the fallback, so will keep it as is.

luchux (approved):

PR #632, opened 2022-01-31
 Create a single context guilds provider #623 

### LINKED ISSUE #623: Create a single context guilds provider
I know this is not specific from your pr, but shouldn't we add one single GuildsContextProvider (maybe after transactionProvider or something like that) instead of adding a new contextProvider per route?

_Originally posted by @MiltonTulli in https://github.com/DXgovernance/dxvote/pull/597#discussion_r794527276_

hhamud:
Closes  #623


mprasanjith (approved):
I feel like we can remove the `GuildsContextProvider`. It only holds `FilterProvider` component that seem to only be used in the Guilds page. Will open an separate issue for this.

PR #641, opened 2022-02-03
Feature/create proposal page

### LINKED ISSUE #605: Create Proposal page layout and inputs
We have a create proposal page with the new editor inside.
However we still need the layout CSS and main inout fields for title and URL as well as the sidebar.
Data in the sidebar has already been implemented and only needs reused via the SWR hooks. 

Screenshot 2022-01-24 at 15 56 58

https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=1399%3A3909

MiltonTulli:
Create proposal page: Edit and preview mode (only UI) 
Closes https://github.com/DXgovernance/dxvote/issues/605

#### Edit mode
Screen Shot 2022-02-03 at 17 56 45

-----------------------------------
#### Preview mode
Screen Shot 2022-02-03 at 17 57 00


src/pages/Guilds/CreateProposal.tsx
```diff
@@ -16,20 +22,169 @@ const SidebarContent = styled(Box)`
...
+    // TODO: add proper validation if toggle from edit to preview without required fields
+    if (editMode && !title && !proposalBodyMd) return;
+    setEditMode(v => !v);
+  };
+
+  const handleBack = () => history.push('/');
+
+  const handleCreateProposal = () => {
+    // TODO: build this functionality
+    console.log('Create proposal click');
```
hhamud: can you remove the console.logs

hhamud (dismissed):
LGTM

src/pages/Guilds/CreateProposal.tsx
```diff
@@ -16,20 +22,169 @@ const SidebarContent = styled(Box)`
...
+  color: ${({ color }) => (color ? color : `#000000`)};
+  margin-bottom: 0.5rem;
+`;
+
 const CreateProposalPage: React.FC = () => {
+  const history = useHistory();
+  const [editMode, setEditMode] = React.useState(true);
+  const [title, setTitle] = React.useState('');
+  const [referenceLink, setReferenceLink] = React.useState('');
+  const [proposalBodyHTML, setProposalBodyHTML] = React.useState('');
```
mprasanjith: I feel like we should move the `useLocalStorageWithExpiry` hooks we currently have in the editor to here, and pass the state down to the editor component as props. Right now we're duplicating the states in both places.

src/pages/Guilds/CreateProposal.tsx
MiltonTulli: Done, thanks!

src/pages/Guilds/CreateProposal.tsx
```diff
@@ -16,20 +22,169 @@ const SidebarContent = styled(Box)`
...
+            </Label>
+          ) : null}
+        </Box>
+        {editMode ? (
+          <Editor
+            onMdChange={setProposalBodyMd}
+            onHTMLChange={setProposalBodyHTML}
+          />
+        ) : (
+          <div dangerouslySetInnerHTML={{ __html: proposalBodyHTML }} />
```
mprasanjith: Not sure about using `dangerouslySetInnerHTML` here. Doesn't seem to be vulnerable at the first glance, but maybe we can render the Markdown instead? The view proposal page is currently rendering Markdown, so it would also bring more parity to how we render the content.

rossneilson: Its fine here so long as we sanitise it, we have more freedom with HTML and its no more dangerous than markdown since markdown is just HTML anyways
https://www.npmjs.com/package/sanitize-html

src/pages/Guilds/CreateProposal.tsx
MiltonTulli: Ok. Refactored to use useLocalStorageWithExpiry hook on create proposal page.

mprasanjith (approved):

PR #651, opened 2022-02-08
docs: New version number 1.0.10

rossneilson:


hhamud (approved):
10% of the way to v1.1

PR #664, opened 2022-02-15
Feature/guilds create proposal

### LINKED ISSUE #631: Create proposal function call
Dependent upon #605 being completed
Now that we have the UI to build at least a signal proposal (one without actions) we can add the actual function call to the smart contracts
The function we want is `createProposal`
We need to first upload ipfs to get the content hash, we don't have a set json data structure yet but it should follow closely to the existing dxvote one but with additional fields for HTML and json versions of the description. 

Eventually this should take the actions selected by the user to send as parameters but for now we can leave this open t later implementation.

mprasanjith:
This wires up the proposal creation workflow to the smart contracts. Fixes #631.

I have also added a draft schema for the IPFS metadata for proposals in `schema/guilds-proposal.json`.

image


src/pages/Guilds/CreateProposal.tsx
```diff
@@ -164,41 +185,14 @@ const CreateProposalPage: React.FC = () => {
           <Button
             onClick={handleCreateProposal}
             variant="secondary"
-            disabled={editMode}
+            disabled={!isValid}
```
MiltonTulli, line 191: Do we want the user to trigger the proposal creation on "edit" mode? I know that in some platforms you need to go to the "preview" mode (if that is the case maybe we should hide this btn on edit mode) Or maybe confirm with Geronimo

src/pages/Guilds/Guilds.tsx
```diff
@@ -42,7 +42,18 @@ const ErrorList = styled(Box)`
 const GuildsPage: React.FC = () => {
   const { chain_name: chainName, guild_id: guildId } =
     useParams<{ chain_name?: string; guild_id?: string }>();
-  const { data: proposalsIds, error } = useGuildProposals(guildId);
+  const { data: proposalIds, error } = useGuildProposals(guildId);
+  const filteredProposals = useMemo(() => {
```
MiltonTulli: Question. Isn't `useGuildProposals` returning just ids? Maybe this can be called `filteredProposalsIds` (and maybe rename the hook too)

rossneilson (dismissed):

MiltonTulli (dismissed):

src/pages/Guilds/Guilds.tsx
mprasanjith: Yeah, you're right, Makes sense to refactor.

rossneilson (dismissed):

hhamud (approved):

PR #665, opened 2022-02-15
Feature/guild styling 658

0xvangrim:
Hello, 

First PR for me. As you can see there are a lot of changed files (most of them are simply styling changes in accordance with @geronimo's Figmas designs: https://www.figma.com/proto/cjLng0AbOKuDj9r5oZVDmR/Wireframe?page-id=3332%3A15606&node-id=3555%3A11926&viewport=388%2C48%2C0.22&scaling=scale-down&starting-point-node-id=3332%3A15607&show-proto-sidebar=1

Here's a summary of the changes:
- Separated our themes to `dark.json` and `light.json` for future toggling of themes.
- Added some frontend logic to the proposal states (might need to change this one depending on how we want to display it) e.g. Active, Passed, Executed, etc.
- Refactored our `Skeleton` to use the `Loading component `instead.
- Refactored our buttons a bit to make use of our Primary and Secondary systems.

These are two of the three views I changed (the third one was if you clicked "Create Proposal" that seemed to have disappeared?). 

**NOTE**: I did not style the information sidebar or the voting sidebar since I got informed that they were being worked on.

Screenshot 2022-02-17 at 18 08 34

Screenshot 2022-02-17 at 18 08 51

Let me know of any changes to make within this PR!


hhamud:
I'm getting a few errors when I try to switch using the network switcher from mainnet into rinkeby.
image

0xvangrim:
> I'm getting a few errors when I try to switch using the network switcher from mainnet into rinkeby. image

Hm yeah. I fixed a small misspelling (my bad) so now you should be able to see the new stylings at least. Regarding the logs, I am getting the same on the `develop` branch so it seems to have been there before this PR I think? 🤔

src/components/Guilds/Filter/index.tsx
```diff
@@ -34,6 +34,8 @@ const ButtonContainer = styled(Flex)`
   flex-direction: Row;
 `;
 
+const SpanContainer = styled.span``;
```
rossneilson: Cant we remove this?

src/components/Guilds/ProposalStatus/index.tsx
```diff
@@ -23,26 +23,27 @@ const Status = styled.div`
...
 `;
 
 const DetailText = styled(Box)`
-  padding: 0 0.2rem;
+padding: 0 0.2rem;
 
-  @media only screen and (min-width: 768px) {
-    padding-right: 0.5rem;
-  }
+@media only screen and(min - width: 768px) {
```
rossneilson: Formatting looks weird here?

rossneilson (dismissed):

src/components/Guilds/ProposalStatus/index.tsx
hhamud: I sometimes also get similiar random formatting errors made by prettier

src/components/Guilds/ProposalStatus/index.tsx
```diff
@@ -84,17 +85,23 @@ const ProposalStatus: React.FC<ProposalStatusProps> = ({
...
     if (!proposal?.endTime) return null;
-
-    if (proposal.state === ProposalState.Submitted) {
-      const currentTime = moment();
-      if (currentTime.isSameOrAfter(proposal.endTime)) {
-        return 'Ended';
-      } else {
-        return 'Active';
-      }
+    switch (proposal.state) {
```
mprasanjith, line 88: Had a chat about this with @Kenny-Gin1. We need to create a separate issue to parse these proposal states properly.

mprasanjith (approved):

mprasanjith:
Regarding the error logs @hhamud is seeing, this happens because the app is trying to connect to the mainnet guild which doesn't exist. So expected behavior, but we should handle it gracefully and show a 404.

mprasanjith (changes requested):
I'm seeing a crash on the proposal editor page. Seems to be related to `polished` library not getting the correct theme values?

0xvangrim:
> I'm seeing a crash on the proposal editor page. Seems to be related to `polished` library not getting the correct theme values?

I think I found the issue and it's related to two files `/components/NetworkModal` and `/components/WalletModal` that refers to theme properties that don't exist. Unfortunately, I don't know the values of those properties nor how to access the page that crashed 😅 . 

Could we mob on this 10 min during the weekend maybe @mprasanjith?

src/components/Guilds/ProposalStatus/index.tsx
rossneilson: Nice, going to make a few tasks soon for detailed specs we need for things like proposal state as well as a few other things to ensure we have one correct util for displaying this

mprasanjith (approved):

PR #671, opened 2022-02-21
Add Comments to Proposal

carlbarrdahl:
This is a proof of concept of adding comments to each proposal using https://theconvo.space
dxdao_comments

### DIFF
--- a/src/hooks/useComments.ts
+++ b/src/hooks/useComments.ts
@@ -1,0 +1,117 @@
+import { DependencyList, useState, useEffect, useCallback } from 'react';
+import { Convo } from '@theconvospace/sdk';
+import { ethers } from 'ethers';
+import { useWeb3React } from '@web3-react/core';
+
+import useJsonRpcProvider from './Guilds/web3/useJsonRpcProvider';
+import { useContext } from '../contexts';
+
+const apiKey = process.env.REACT_APP_CONVO_API_KEY;
+const spaceUrl = window.location.origin;
+const convo = new Convo(apiKey);
+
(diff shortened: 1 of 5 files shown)


src/hooks/useComments.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+      return resp.message;
+    } catch (error) {
+      console.log(error);
+    }
+  }, [provider]);
+}
+
+// Helper functions.
+// Would suggest to use react-query or react-use instead to help with async state management
```
mprasanjith, line 77: We're starting to adopt `swr`, which might be useful here?

hhamud:
Run yarn format as well when you've finished implementing the changes that @mprasanjith  has recommended. It's also best that we merge #602 before this as one as @AugustoL has refactored the cache to use SWR

carlbarrdahl:
Thanks for your review @mprasanjith and @hhamud!

SWR is also a good choice!
Here's a comparison: https://react-query.tanstack.com/comparison

This was mostly a proof of concept I did for the EthDenver hackathon (bounty 5). I wasn't able to submit to the hackathon and thought I would at least send a PR here. 

Before merging maybe there should be a discussion if this is a feature that we want included and if it's the correct implementation (to use Convo.space).

rossneilson:
Likely time we can close this. Although we didn't end up merging this was an amazing proof of concept and most excitingly our first open source contribution. 
It also led to Carl joining the DXgov team, prototyping and now building an awesome off chain social layer for governance. 
Time taken to build this was taken into account for his trial period's work.

PR #676, opened 2022-02-27
Refactor/dao store methods

AugustoL:


src/stores/DaoStore.ts
```diff
@@ -426,13 +383,13 @@ export default class DaoStore {
     };
   }
 
-  getAllProposals(): ProposalsExtended[] {
+  getAllProposals(filter: any = {}): ProposalsExtended[] {
```
hhamud, line 386: Could we place a specific type here?.

src/stores/DaoStore.ts
```diff
@@ -446,36 +403,71 @@ export default class DaoStore {
...
+        ],
+    };
   }
 
-  getVotingParametersOfScheme(schemeAddress): VotingMachineParameters {
-    const scheme = this.daoCache.schemes[schemeAddress];
-    return this.daoCache.votingMachines[scheme.votingMachine].votingParameters[
-      scheme.paramsHash
-    ];
+  getVotingMachineOfProposal(proposalId): {
```
hhamud: Can we specify the proposalId type?

src/hooks/useContract.ts
```diff
@@ -16,4 +18,22 @@ export default function useContract(address, abi) {
...
+  functionName: string,
+  params: string[]
+): string => {
+  const { active, library } = useWeb3React();
+  const { data } = useSWR(id(address + functionName + params), async () => {
+    if (!active) return '0x0';
+
+    const provider = new providers.Web3Provider(library.currentProvider);
+    const tokenContract = getContract(address, abi, provider);
+    return await tokenContract[functionName](...params);
```
hhamud, line 35: I don't understand what you are trying to do here exactly?

hhamud (commented):
I've left behind a few comments but the rest of the code looks fine to me. I think we also could probably refactor the useContract to enable using typechain interfaces.

src/stores/DaoStore.ts
AugustoL: I think it is not necessary, since we just using it as a filter for the proposals array, we can leave it as any.

src/hooks/useContract.ts
AugustoL: A hook to do contract call, with the address, abi, function name and funtion params.

The value fetched is saved in the SWR cache with a deterministic id, the value fetched is saved on the SWR cache on the id with the hash of the contract address, function and params, which means that the only way to get the stored value is doing the same call to the same smart contract again.

AugustoL: Later it would be good to see how we can use typechain contracts.

src/stores/DaoStore.ts
AugustoL: https://github.com/DXgovernance/dxvote/pull/676/commits/8fa43c8f80607c8485ef7f250bf19f7a2aa76489

src/hooks/useContract.ts
hhamud: ok, I understand now. `id(..)` is just hashing the provided parameters. we can do the refactoring of this at a later date if you wish so.

hhamud (approved):

PR #684, opened 2022-03-03
Add Action Modal for the Actions Widget

### LINKED ISSUE #683: Add Action Modal for the Actions Widget
Before you read the whole description I encourage you to play around with the playground [here](https://www.figma.com/proto/cjLng0AbOKuDj9r5oZVDmR/Wireframe?page-id=4442%3A17030&node-id=4442%3A17031&viewport=325%2C48%2C0.06&scaling=min-zoom&starting-point-node-id=4442%3A17031)!

One of the main features of the `Actions widget` is to allow the user to edit vote options and its individual action(On-chain) package. A user can do that by toggling the edit button the header of the actions widget:
Desktop
[Figma design specification](https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=1959%3A6979)

Once in the edit mode, under each option a button appears with the description `Add action` which opens up the `Add Action` Modal:

AddAction

[Figma design specification](https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=1767%3A6400)

- If the `Actions Widget`already has the `Transfer & Mint` component added, the `Transfer & Mint` Button under the subheader `Core` should be disabled.
- Under the subheader `external contracts` should all whitelisted contracts be listed with a counter of function calls (we call it here actions).
- User can close the modal by clicking on the close icon in the top right corner of the modal.
- Once the user decides which external contract the DAO should interact with we are gonna show a list of all actions available:
DXdaoController

[Figma design specification](https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=3241%3A12074)
- If a user wants to go back to change the external contract it should be able to click on `< DXdaoController`
- User can close the modal by clicking on the close icon in the top right corner of the modal.
- This modal lists all functions/actions the DAO can interact with the external contract.
- Once the user decides which functions we are transitioning to:
MintReputation

[Figma design specification](https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=3241%3A13508)
- the function parameters define the last modal. If the function/action requires an address we show an address inputfield. If a number we show a number inputfield etc.
- If a user wants to go back to change the external contract it should be able to click on `< Mint Reputation`
- Once the users fills out all the inputfields the `Add Action` button is active and the user can proceed with adding it to the `actions widget`:
Desktop

[Figma design specification](https://www.figma.com/file/cjLng0AbOKuDj9r5oZVDmR/Wireframe?node-id=2027%3A5784)

- For now all external function calls will be shown with the code icon followed by the function name and the function(parameters).
- Users can edit the action by clicking edit which will reopen the Add action modal.
- Users can collapse the action to only see the function call, the Default view is disabled(might get added later).

hhamud:
* Closes  #683 
 ### Notes: 
* I had added a temporary boolean to disable the add action button so as to prevent the `TransferAndMint` component from moving into the other actions component as the way the contract data will be handled hasn't been decided yet.
* I threw away the useReducer refactoring and kept it all the state change handling with `useState` for the similar above reason.


src/components/Guilds/CreateProposalPage/ActionsBuilder/index.tsx
```diff
@@ -101,7 +105,11 @@ export const ActionsBuilder: React.FC<ActionsBuilderProps> = ({ editable }) => {
             key={idx}
             data={option}
             editable={editable && actionsEditMode}
+            showData={mintRep}
           />
+          {transferBuilder && editable && option.label === 'For' && (
```
rossneilson: This isnt really how guild's competition style of voting is going to work as the against and all other options can have transfer actions also

(at least in the base implementation - we will support a required blank against later)

src/components/Guilds/ActionsModal/DXDController.tsx
```diff
@@ -0,0 +1,67 @@
+import styled from 'styled-components';
```
rossneilson: Obviously we dont yet have the configs ready to consume so I understand why its not built like that yet and is mostly hardcoded but we definitely shouldn't be baking in DXDController into actions modal components. This is a product that many teams externally to DXdao will use and so we need to make as much of it as possible configurable and avoid hardcoding like this. 
Instead this could be a Contract component representing one contract and all the calls that can be made to it, we could then pass it as harcoded props for now.

rossneilson (changes requested):
Nice this is a good start on implementing all the styles required and I like the actions builder hook.
Just a few things that arent really in the style we want to be building in although understandable since the configs arent available to consume yet

src/components/Guilds/CreateProposalPage/ActionsBuilder/index.tsx
hhamud: I wasn't sure how exactly this was going to look like so I placed that simply to view the css component. I'll refactor this then.

src/components/Guilds/ActionsModal/DXDController.tsx
hhamud: Good Point, I'll make this more flexible.

rossneilson (approved):

PR #686, opened 2022-03-04
Feature/update proposal page actions components

### LINKED ISSUE #679: Update proposal page actions components
We haven't updated the proposal page action display component for quite a while and as a result its out of sync with figma.

existing implementation
Screenshot 2022-02-28 at 12 58 48

We shouldreuse the component from the action builder since they are exactly the same only in this case it cant be edited. 

New design
Screenshot 2022-02-28 at 12 59 05

For now the placeholder data can be kept also

mprasanjith:
Implements the proposal page Actions viewer. This mostly re-uses the same ActionsBuilder component we're using in the proposal editor. Closes #679.

Because we haven't defined the schema for keeping the action details, I used the same structure we use for RecommendedCalls in DXvote with some changes, but we'll have to change this later probably.

image

image

image


src/components/Guilds/CreateProposalPage/ActionsBuilder/Option/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  font-weight: 500;
+  font-size: 14px;
+  line-height: 20px;
+  color: ${({ theme }) => theme.gray};
+`;
+
+const OptionRow: React.FC<OptionRowProps> = ({ data, editable }) => {
+  return (
+    <Box padding="16px">
+      <Flex padding="8px 0" direction="row" justifyContent="space-between">
```
hhamud: better to create wrapper components

src/components/Guilds/CreateProposalPage/ActionsBuilder/ProposalOptionTag/ProposalOptionTag.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const Tag = styled.span<ProposalActionTagProps>`
+  margin: 0;
+  border-radius: 6px;
+  padding: 4px 8px;
+  border: 1px solid ${({ theme }) => theme.colors.muted};
+  font-style: normal;
+  font-weight: 600;
+  font-size: 14px;
+  line-height: 20px;
```
hhamud: is it not possible to use themes with this?

src/components/Guilds/CreateProposalPage/ActionsBuilder/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+import SidebarCard, {
+  SidebarCardHeaderSpaced,
+} from 'components/Guilds/SidebarCard';
+import OptionRow, { Option } from './Option';
+import AddButton from './AddButton';
+
+const Button = styled(CommonButton)`
+  font-style: normal;
+  font-weight: 600;
+  font-size: 12px;
```
hhamud: themes?

src/components/Guilds/CreateProposalPage/ActionsBuilder/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+            editable={editable && actionsEditMode}
+          />
+          {idx !== options.length - 1 && <Divider />}
+        </>
+      ))}
+
+      {editable && actionsEditMode && (
+        <>
+          <Divider />
+          <Box padding="16px">
```
hhamud: I think it's better to wrap the component around such that we won't be tempted to just add further changes inline.

hhamud (commented):
Just a few changes that are needed before merging

src/components/Guilds/CreateProposalPage/ActionsBuilder/Option/index.tsx
mprasanjith: We should get rid this of this `Flex` component which promotes inline styling at some point.

mprasanjith:
@hhamud Apologies for the inline styles. I made a mental note to refactor them and then completely forgot. Hope this is okay now?

hhamud (approved):
LGTM!

hhamud:
> @hhamud Apologies for the inline styles. I made a mental note to refactor them and then completely forgot. Hope this is okay now?

No need for apologies. Your code is always good and I refer to them as a general example to follow.

PR #687, opened 2022-03-07
feat/develop-script

AugustoL:
Changes the behavior of the `yarn start` and `yarn dev` commands.

Now yarn start would do a production build, place it in the build folder and start a local http-server on the port 8080 serving the files in the build folder (long process but this is the closer we can get locally to a production env)

`yarn dev` will run the deploy-dxvote hardhat task imported form dxdao-contracts repository with a script placed in the scripts/dev.ts file, the developer has complete control over whats deployed and what happens before the dapp starts, once the hardhat node starts the contracts are deployed, and actions executed. Then the app starts with `yarn react-scritps start` on development environment.

An  bug were  the finalized proposals not being shown in the proposal table https://github.com/DXgovernance/dxvote/pull/687/commits/ee1c271aafd5031b73f27c292e60eef8ce9ddfb9 was fixed too.

Also the useContracts hook for dxvote now uses the ETH-SWR multicall, and it was added in the header to fetch the user balances.

Also, typechain was added to hardhat, so everytime you do `hardhat compile` it will generate the typings and place it in src/types/contracts.


scripts/dev.sh
```diff
@@ -84,24 +82,25 @@ start-hardhat_node() {
...
 }
 
-if ganache_running; then
-  echo "Using existing hardhat node instance"
-else
-  echo "Starting our own hardhat node instance"
-  start-hardhat_node
+if hardhat_running; then
+  echo "Killing existent hardhat"
+  fuser -k 8545/tcp
```
MiltonTulli: Fuser is only Linux (I believe?). 
Something like this should work for both Linux/macOS
```
kill $(lsof -t -i:8545) 
```

scripts/dev.ts
```diff
@@ -0,0 +1,542 @@
...
+          'https://s2.coinmarketcap.com/static/img/coins/200x200/5589.png',
+      },
+    ],
+    guilds: [addresses.DXDGuild, addresses.REPGuild],
+  };
+
+  await writeFileSync(
+    './src/configs/localhost/config.json',
+    JSON.stringify(developConfig, null, 2)
+  );
```
MiltonTulli: We can be sure that the folder exists, if not this will fail. 
like:
```
  mkdirSync(path.resolve(__dirname, '../src/configs/localhost'), {
    recursive: true,
  });
  await writeFileSync(
    path.resolve(__dirname, '../src/configs/localhost/config.json'),
    JSON.stringify(developConfig, null, 2)
  );
```

AugustoL: done in https://github.com/DXgovernance/dxvote/pull/687/commits/b0cca80c4b9fc3a521958523e6e6bcc46749023d

scripts/dev.sh
AugustoL: done in https://github.com/DXgovernance/dxvote/pull/687/commits/b0cca80c4b9fc3a521958523e6e6bcc46749023d

MiltonTulli (approved):

PR #688, opened 2022-03-07
Feature/vote summary

### LINKED ISSUE #659: Display votes for proposals
Currently, all votes are just hardcoded placeholders. We should replace this with real data now.
We actually already have this data from fetching `getProposal` under `totalVotes`. This value corresponds to the actions of a proposal and so the array will have an equal number of elements as actions for a proposal. This can then be used to calculate the number of votes for each action.
For now the total number of votes available to divide this value by is just the total number of tokens locked in a guild. However we should add a flag as a parameter which can use snapshot style voting. For now we don't need to implement the logic for this as we would need a deployed snapshot guild to test but it is good to have it as an option for future.

rossneilson:
Closes #659 
Displays votes in proposal cards and adds a few utilities
Limiting displayed votes to 3 and currently, instead of looping through an array I set the loop to 2 because we are not enforcing at least 1 action on proposal creation. Also added a comment to fix that.

### DIFF
--- a/src/components/Guilds/ProposalCard/index.tsx
+++ b/src/components/Guilds/ProposalCard/index.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react';
 import styled from 'styled-components';
 import { useParams } from 'react-router';
 import { isDesktop } from 'react-device-detect';
@@ -14,6 +15,7 @@ import Avatar from '../Avatar';
 import { DEFAULT_ETH_CHAIN_ID } from '../../../provider/connectors';
 import { shortenAddress } from '../../../utils';
 import { Loading } from '../common/Loading';
+import useVoteSummary from 'hooks/Guilds/useVoteSummary';
 
 const CardWrapper = styled(Box)`
   border: 1px solid ${({ theme }) => theme.colors.muted};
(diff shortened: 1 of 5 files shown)


src/components/Guilds/ProposalCard/index.tsx
```diff
@@ -109,10 +111,35 @@ interface ProposalCardProps {
...
+  const [votesSummary, setVoteSummary] = useState([]);
+
+  // Build vote summary array
+  useEffect(() => {
+    const voteSummaryArray = [];
+
+    votes
+      .sort((a, b) => b - a)
+      .map((vote, i) => {
+        if (i < 3 && !(i === votes.length - 1)) {
```
mprasanjith: Maybe we should just set only the data and not the whole component to the state for performance reasons. Also, I think we can use `useMemo` instead of `useEffect` here.

What are we actually going to show if there are more than 2 options? Maybe something to discuss with Gero.

src/hooks/Guilds/useVoteSummary.ts
```diff
@@ -0,0 +1,37 @@
...
+  guildId: string,
+  proposalId: string
+): number[] {
+  const [votes, setVotes] = useState<number[]>([]);
+  const { data: { totalVotes = null } = {} } = useProposal(guildId, proposalId);
+  const { data: { totalLocked = null } = {} } = useGuildConfig(guildId);
+
+  useEffect(() => {
+    if (totalVotes && totalLocked) {
+      const getVoteData = async () => {
```
mprasanjith: Any reason why this is async? If we don't need this to be async we can memoize this nicely.

mprasanjith (commented):
This looks great! Just left a couple of comments with potential performance improvements maybe we can handle before merging.

mprasanjith (approved):

PR #689, opened 2022-03-07
docs: Adding schema for contract configuration

### LINKED ISSUE #685: Guild contract config schema
The idea is that we will be able to curate a collection of configs for guild contract interactions that is owned and curated by dxdao
(we would of course add override functionality here its just that it would break the UX for others as there would be no human readable description)
We dont want etherscan reliance but we absolutely need generic external contracts and something similar to recommended calls is the best way.

This schema must define the fields needed to decode and show readable descriptions as well as allow users to create actions for this proposal in an intuitive way

rossneilson:
Closes #685 
Adding schema for contract configuration
This is an initial idea for a schema we can use to make a store of modular contract configs which guilds and daos can choose from a registry / collection curated by dxdao
This should give the basic details needed to encode and decode calls to contracts as well as allowing for configuration that should allow for a good UX without too many technical terms.

### DIFF
--- a/schema/contract-config.json
+++ b/schema/contract-config.json
@@ -1,0 +1,36 @@
+{
+  "$id": "http://example.com/example.json",
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "default": {},
+  "description": "The root schema comprises the entire JSON document for contract configuration schema.",
+  "required": [
+      "title",
+      "networks",
+      "functions"
+  ],
+  "title": "Vesting contract for DXD token",
+  "networks": {
+      "mainnet": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "rinkeby": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "gnosis": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "arbitrum": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "arbitrumRinkeby": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B"
+  },
+  "functions": [
+      {
+          "title": "Create vesting contract",
+          "functionName": "create(address, uint256, uint256, uint256, uint256)",
+          "params": [
+              { "type": "address", "component": "address", "name": "beneficiary", "defaultValue": "", "description": "Receiving address of tokens" },
+              { "type": "uint256", "component": "date", "name": "start", "defaultValue": "", "description": "Starting time for contract" },
+              { "type": "uint256", "component": "time", "name": "cliffDuration", "defaultValue": "", "description": "How long before cliff" },
+              { "type": "uint256", "component": "time", "name": "duration", "defaultValue": "", "description": "How long before contract finishes fully" },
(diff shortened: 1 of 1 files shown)


schema/contract-config.json
```diff
@@ -0,0 +1,36 @@
...
+      "mainnet": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "rinkeby": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "gnosis": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "arbitrum": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B", 
+      "arbitrumRinkeby": "0x0b17cf48420400e1D71F8231d4a8e43B3566BB5B"
+  },
+  "functions": [
+      {
+          "title": "Create vesting contract",
+          "functionName": "create(address, uint256, uint256, uint256, uint256)",
```
mprasanjith: Maybe this could be just the function name (just `create` instead of `create(address, uint256, uint256, uint256, uint256)`. We can always get the param types from the params array. Otherwise it becomes complicated to format it the way we want:

image

schema/contract-config.json
```diff
@@ -0,0 +1,36 @@
...
+          "params": [
+              { "type": "address", "component": "address", "name": "beneficiary", "defaultValue": "", "description": "Receiving address of tokens" },
+              { "type": "uint256", "component": "date", "name": "start", "defaultValue": "", "description": "Starting time for contract" },
+              { "type": "uint256", "component": "time", "name": "cliffDuration", "defaultValue": "", "description": "How long before cliff" },
+              { "type": "uint256", "component": "time", "name": "duration", "defaultValue": "", "description": "How long before contract finishes fully" },
+              { "type": "uint256", "component": "tokenAmount",  "name": "value", "defaultValue": "", "description": "Number of tokens to vest" }
+          ],
+          "shortDescription": "",
+          "longDescription": "",
```
mprasanjith: Maybe `longDescription` could be markdown? Just thinking out loud.

mprasanjith (changes requested):
Two concerns:

1. Don't think this is valid json-schema. I usually use [jsonschema.net](https://www.jsonschema.net/) to quickly generate the correct schema from data.

2. How are we going to generate this simplified view?
image

rossneilson:
> How are we going to generate this simplified view?

So this view specifically will be a little more special in the same way that ERC20 transfer calls will be special also.
This schema is meant to show the generic calls so anything beyond erc20 and rep

rossneilson:
> 1. Don't think this is valid json-schema. I usually use [jsonschema.net](https://www.jsonschema.net/) to quickly generate the correct schema from data.

I was copying from the existing schema we have in the repo but it seems it also may not be valid

schema/contract-config.json
rossneilson: Good idea!

schema/contract-config.json
rossneilson: It certainly could be but I think keeping it as plain text gives us a little more freedom in how we use it.
I was mostly imagining long text being used for things like on hover help information for users

mprasanjith (approved):

PR #691, opened 2022-03-07
Feature/guilds error pages

### LINKED ISSUE #654: Main page error display
We should have some error handling for the main page of the application.
We don't yet have designs for these but for now they can just be an unstyled message.

We should show two main errors here, one for the guild not existing (i.e. when you connect to mainnet and its only on test networks). This can be checked with the deployed() function 
https://docs.ethers.io/v5/api/contract/contract/#Contract-deployed
If you want brownie points you could also show a list of networks the contract is deployed on since it will have the same address.

And also and error for when no proposals are found.

mprasanjith:
- Adds error messages to most of the possible error scenarios. Fixes #654.
- Generic `Result` component has been added with support for Info / Warning / Error / Success states. Inspired by how [Ant Design](https://ant.design/components/result/) implements the same component.
- Guild contract availability checks have been added using a global provider.
- Network switching functionality has been moved to its own hook.
- `browserslist` database has been updated to the latest version.

Guild unavailable on active network

Incorrect guild address

Incorrect proposal ID

Generic 404


src/components/Guilds/common/Result/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+      {icon ||
+        createElement(IconMap[state], {
+          size: 32,
+          color:
+            state === ResultState.ERROR
+              ? theme?.colors?.red
+              : theme?.colors?.primary,
+        })}
+      <Title size="2">{title}</Title>
+      <Subtitle size="1">{subtitle}</Subtitle>
```
hhamud: is it not possible to include the size props within the styled.css component?

src/components/Guilds/AvailabilityGuard/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    '/:chain_name/:guild_id'
+  );
+  const { chainId } = useWeb3React();
+
+  const guildId = routeMatch?.params?.guild_id;
+  const availability = useContractAvailability(guildId);
+
+  if (
+    Object.keys(availability).includes(String(chainId)) &&
+    !availability?.[chainId]
```
hhamud: It may be better to encapsulate this within its own variable. e.g. ```isGuildAvailable```

src/contexts/Guilds/guildAvailability.tsx
```diff
@@ -0,0 +1,126 @@
...
+        }
+      />
+    );
+  }
+
+  return (
+    <GuildAvailabilityContext.Provider
+      value={{
+        availability,
+        isLoading: !Object.keys(availability).includes(String(currentChainId)),
```
hhamud: its also repeated here so I think encapsulating this would be better

hhamud (commented):
Great work, left a few comments on the code.

src/components/Guilds/common/Result/index.tsx
mprasanjith: Didn't realize this is a feature!

hhamud (approved):
Awesome

PR #699, opened 2022-03-16
Bugfix/frame network switching fixes

rossneilson:
Added code to handle switching via network modal for frame


mprasanjith (approved):

rossneilson:
There is still an issue with frame not firing chainChanged events properly causing switching directly in the wallet to not update the UI but for now at least it works via the network switcher

PR #795, opened 2022-04-14
Refactor Cache Service

AugustoL:
# Description

- Add processProposal that gets the immutable + mutable data if the proposal is new, and if not it gets only the mutable data and updates the proposal in the cache.
- Add processScheme fucntion that gets the immutable + mutable data if the scheme is new, and if not it gets only the mutable data and updates the scheme in the cache. Also if the scheme is new it gets the proposal creation events and adds the new proposals.
- Refactor code form cacheService to utils, adding more utils function and removing duplicated code.
- Removed some unnecesary varibales requested as parameters.

I tested the xdai cache generated againts the latest we build for xdai and there are very few differences, in fact, the cache built from this PR is more up to date.

## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/src/services/CacheService.ts
+++ b/src/services/CacheService.ts
@@ -822,20 +822,19 @@ export default class UtilsService {
     await batchPromisesOntarget(
       Object.keys(networkCache.proposals)
         .filter(proposalId => {
+          const proposal = networkCache.proposals[proposalId];
+          const scheme = networkCache.schemes[proposal.scheme];
           return (
             // Only update proposals for registered schemes
-            networkCache.schemes[networkCache.proposals[proposalId].scheme]
-              .registered &&
+            scheme.registered &&
             // This condition check that the proposal already existed in the current block range.
             // If not, it means that the proposal was created in the current block range when processing the scheme,
             // So there is no need to process it again.
-            networkCache.proposals[proposalId].creationEvent.blockNumber <
-              fromBlock &&
+            proposal.creationEvent.blockNumber < fromBlock &&
             // This condition check that the proposal is active
-            (networkCache.proposals[proposalId].stateInVotingMachine >
+            (proposal.stateInVotingMachine >
               VotingMachineProposalState.Executed ||
-              networkCache.proposals[proposalId].stateInScheme ===
-                WalletSchemeProposalState.Submitted)
+              proposal.stateInScheme === WalletSchemeProposalState.Submitted)
           );
         })
         .map(proposalId => {
--- a/src/services/CacheService.ts
+++ b/src/services/CacheService.ts
@@ -1,5 +1,12 @@
 import RootContext from '../contexts';
-import { batchPromisesOntarget, getIPFSFile, NETWORK_NAMES } from '../utils';
+import {
+  batchPromisesOntarget,
+  getAppConfig,
+  getDefaultConfigHashes,
+  getIPFSFile,
+  getProposalTitles,
+  NETWORK_NAMES,
+} from '../utils';
 import Web3 from 'web3';
 import _ from 'lodash';
 import {
@@ -23,26 +30,6 @@ import { getContracts } from '../contracts';
 
 const Hash = require('ipfs-only-hash');
 const jsonSort = require('json-keys-sort');
-
-const defaultConfigHashes = require('../configs/default.json');
-
-const arbitrum = require('../configs/arbitrum/config.json');
-const arbitrumTestnet = require('../configs/arbitrumTestnet/config.json');
-const mainnet = require('../configs/mainnet/config.json');
-const xdai = require('../configs/xdai/config.json');
-const rinkeby = require('../configs/rinkeby/config.json');
-const localhost = require('../configs/localhost/config.json');
-
-const proposalTitles = require('../configs/proposalTitles.json');
-
-const appConfig: AppConfig = {
-  arbitrum,
-  arbitrumTestnet,
-  mainnet,
-  xdai,
-  rinkeby,
-  localhost,
-};
 
 export default class UtilsService {
   context: RootContext;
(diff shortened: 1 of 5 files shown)


src/services/CacheService.ts
```diff
@@ -711,35 +668,65 @@ export default class UtilsService {
...
-            start: callsResponse.decodedReturnData[4],
-            isOwner: callsResponse.decodedReturnData[5],
-            revocable: callsResponse.decodedReturnData[6],
+            beneficiary: callsResponse.decodedReturnData[0][0],
+            cliff: callsResponse.decodedReturnData[1][0],
+            duration: callsResponse.decodedReturnData[2][0],
+            owner: callsResponse.decodedReturnData[3][0],
+            start: callsResponse.decodedReturnData[4][0],
+            isOwner: callsResponse.decodedReturnData[5][0],
+            revocable: callsResponse.decodedReturnData[6][0],
```
MiltonTulli, line 730: I think we are not using these 2 keys (`isOwner` `revocable`) in dxvote and we can avoid storing them in the cache unless we know that we will use them in the closer future.

src/services/CacheService.ts
```diff
@@ -758,1318 +745,116 @@ export default class UtilsService {
...
+            // If not, it means that the proposal was created in the current block range when processing the scheme,
+            // So there is no need to process it again.
+            networkCache.proposals[proposalId].creationEvent.blockNumber <
+              fromBlock &&
+            // This condition check that the proposal is active
+            (networkCache.proposals[proposalId].stateInVotingMachine >
+              VotingMachineProposalState.Executed ||
+              networkCache.proposals[proposalId].stateInScheme ===
+                WalletSchemeProposalState.Submitted)
+          );
```
MiltonTulli: Just thinking of making the cacheService more readable and debuggable, I suggest:
```suggestion
    const proposal = networkCache.proposals[proposalId];
    const scheme = networkCache.schemes[proposal?.scheme];
    return (
      // Only update proposals for registered schemes
      scheme?.registered &&
      // This condition check that the proposal already existed in the current block range.
      // If not, it means that the proposal was created in the current block range when processing the scheme,
      // So there is no need to process it again.
      proposal?.creationEvent.blockNumber < fromBlock &&
      // This condition check that the proposal is active
      (proposal?.stateInVotingMachine > VotingMachineProposalState.Executed ||
        proposal?.stateInScheme === WalletSchemeProposalState.Submitted)
    );
```

src/services/CacheService.ts
```diff
@@ -2150,4 +935,957 @@ export default class UtilsService {
...
+  // If the proposal does not exist in the cache it gets the immutable + mutable data
+  // The proposal already exists it gets the mutable data only
+  async processProposal(
+    proposalId: string,
+    networkCache: DaoNetworkCache,
+    networkContracts: NetworkContracts,
+    web3: Web3,
+    fromBlock: number,
+    toBlock: number,
+    creationEvent?: any
```
MiltonTulli: Should we add types for this? (BlockchainEvent?)
```
{
event,
signature,
address,
transactionHash,
blockNumber,
timestamp,
transactionIndex,
logIndex,
}

File: src/utils/helpers.ts [hunk 4]
```diff
@@ -87,3 +87,20 @@ export const appendEthAPIKey = (
 export function escapeRegExp(string: string): string {
   return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
 }
+
+export async function retryPromise(
```
MiltonTulli, line 91: Should we add a `maxTries` arg?

MiltonTulli (dismissed):
Didn't test it but the code looks good to me!. Just some comments.

src/services/CacheService.ts
AugustoL: https://github.com/DXgovernance/dxvote/pull/795/commits/ecf96b5ea0948d9d1ea208dc8148941cf013f97b

src/utils/helpers.ts
AugustoL: https://github.com/DXgovernance/dxvote/pull/795/commits/97e931ccfee1013cce4e716dd240c662c533eb18

src/services/CacheService.ts
AugustoL: https://github.com/DXgovernance/dxvote/pull/795/commits/06704c688de6d7413117543c2bffce88574bada3

src/services/CacheService.ts
AugustoL: No need to change this now, this is one of the first things that can be moved to hooks from the cache context, also we might want that fields in a future.

rossneilson:
Testing this I am seeing some pretty large differences from develop, both loaded at the same time from remote cache
1 boosted vs 4 boosted
0 pre boost vs 5 pre boost
2 queued vs 4 queued
Looks like the missing ones don't exist
Preview:
Screenshot 2022-04-21 at 12 53 26
Develop:
Screenshot 2022-04-21 at 12 53 44

AugustoL:
> Testing this I am seeing some pretty large differences from develop, both loaded at the same time from remote cache 1 boosted vs 4 boosted 0 pre boost vs 5 pre boost 2 queued vs 4 queued Looks like the missing ones don't exist Preview: Screenshot 2022-04-21 at 12 53 26 Develop: Screenshot 2022-04-21 at 12 53 44

hmmm, going to debug, thx

AugustoL:
@rossneilson should be fixed now. I comapred a resetted cache and only mutable values changed in some proposals, mainly boosts and bounty redeems.

rossneilson (approved):

PR #796, opened 2022-04-18
v1.0.14

AugustoL:
# Description

dxvote v1.0.14


rossneilson:
Dont think we want to merge from your develop into master and don't think we have anything we are releasing soon without some more QA around cache changes?
Shall we close this?

AugustoL:
ups, yes it was supposed to be form the develop branch here. We could have had it opened anyway, just to follow up what commits are going to be merged (after changing the develop branch)

PR #802, opened 2022-04-21
Feature/connection loader

AugustoL:
# Description

I think it was my bad internet conenction that was causing a white screen to stay for few seconds before the app showed anything, turned out it was that we were returning null while trying to eager connect, I think the eager connect might not be a good idea, we shoudl investigate more if it is still being used.

Anywayyyyy I fixed by returning a loading skeleton of the dapp instead, now at least I see a beatiful pulting loading icon with the DXD logo while waiting for the app to load instead of a white screen.

Closes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/src/components/Header/loadingNetwork.tsx
+++ b/src/components/Header/loadingNetwork.tsx
@@ -1,0 +1,51 @@
+import { observer } from 'mobx-react';
+import styled from 'styled-components';
+import dxdaoIcon from 'assets/images/DXdao.svg';
+
+const NavWrapper = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  width: 100%;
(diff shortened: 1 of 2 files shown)


mprasanjith:
Now that you mention it, the reason why it was showing a white screen when using Frame seems to also be eager connect. Didn't connect the dots before. It's a nice feature, although different wallets seem to react to it differently. If we continue having issues we could remove it.

mprasanjith (approved):

rossneilson (approved):

PR #817, opened 2022-04-27
Fix/proposal card test

MiltonTulli:
# Description
Add missing key prop in mapping components inside ProposalCard component to avoid log error on each commit (from test script)

Before: 
Screen Shot 2022-04-27 at 11 39 36

After: 
Screen Shot 2022-04-27 at 11 49 22


rossneilson:
This change alters the actual behaviour of the component

PR #846, opened 2022-05-10
Feature/string interpolation generic calls

### LINKED ISSUE #804: Use string interpolation to display generic calls
In rich contract config we have a `templateLiteral` value in the function object that describes how to display the description of a proposal call with the values shown in context.
This follows the same style as javascript with ${parameterOne} describing the values to be replaced. These values are expected to match exactly the parameter names if the function.

We will also need to look up the type of the value to properly format it, the best way will be referring to the `component` value, for example: time, address, tokenAmount, etc

This will be displayed exactly where we currently show just the raw function signature, but keep this raw function signature implementation as we will need it as a backup option later. 
Screenshot 2022-04-21 at 11 37 50

mprasanjith:
# Description

- Refactor the `useDecodedCalls` hook results to include the rich contract registry data.
- Implement showing templated details in generic call actions, and adds a few components to support templating.
- Rename `ContractRegistry` related functions and files to `RichContractData` to be consistent with the name we're using.
- Introduce tests for `useDecodedCalls` and `useRichContractData` hooks.
- Improve tests for `Avatar` components.

Closes #804.

## Type of change

# How Has This Been Tested?

Tested locally, as well as using unit tests (Partially. More unit tests will need to be added in the future.)

# Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas


src/hooks/Guilds/contracts/useDecodedCall.test.ts
```diff
@@ -0,0 +1,68 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { BigNumber } from 'ethers';
+import { SupportedAction } from 'old-components/Guilds/ActionsBuilder/types';
+import { ZERO_ADDRESS, ZERO_HASH } from 'utils';
+import { useDecodedCall } from './useDecodedCall';
+
+const mockChainId = 1337;
```
rossneilson: Could we set mockChainId to a non real id just to avoid confusion?

src/hooks/Guilds/contracts/useDecodedCall.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  useWeb3React: () => ({
+    chainId: mockChainId,
+  }),
+}));
+
+jest.mock('./useRichContractData', () => ({
+  useRichContractData: () => ({
+    contracts: [],
+  }),
+}));
```
rossneilson, line 22: We should add a test for decoding via rich contract data also but we can do this later

src/hooks/Guilds/contracts/useDecodedCallRichContractData.ts
```diff
@@ -0,0 +1,32 @@
+import { DecodedCall } from 'old-components/Guilds/ActionsBuilder/types';
+import { useMemo } from 'react';
+import { RichContractFunction } from './useRichContractData';
+
+const useDecodedCallRichContractData = (decodedCall: DecodedCall) => {
```
rossneilson: What do you think of `useRichContractData` here and then maybe `useRichContractConfig` for the hook that fetches from ipfs, the names are a little complex like this I think
Definitely a good decision to use RichContract everywhere though, nice job

src/old-components/Guilds/ActionsBuilder/SupportedActions/GenericCall/GenericCallParamsMatcher.tsx
```diff
@@ -0,0 +1,78 @@
...
+      case 'contentHash':
+      default:
+        return param.value;
+    }
+  }
+
+  match(string: string): MatchResponse<MatchResult> | null {
+    const params = this.options.params;
+
+    console.log({
```
rossneilson: Console log

src/hooks/Guilds/contracts/useDecodedCall.test.ts
mprasanjith: Yeah, we need to add more tests to test all the individual functions (`decodeCallUsingEthersInterface`, `getContractInterfaceFromRichContractData`, `getContractFromKnownSighashes` etc.). I didn't do it in this PR as it's a bit complicated to build the input data and mock everything. We should tackle that soon though.

rossneilson (approved):

Issue #755, opened 2022-03-28
Action Builder: Use Sourcify to resolve ABIs


mprasanjith:
We want the second stage of contract details to try to pull the ABI from sourcify (if not present in rich contract config)
Later we will use this for encoding but since we currently build the generic calls list from rich contract data there is not much to do here for now. Later we will add in a custom option which will reuse the sourcify abi hooks for encoding. 

**Acceptance criteria:**
- When decoding actions check first for rich-contract data and if none is found then check sourcify using partial match.
- Show as much data as we can from the ABI. If it includes natspec comments we can use these as descriptions, if no comments then we will just show the method name and parameter name and types
- If no sourcify match is found then display the unknown contract message that already exists
- Test by finding a contract that exists in sourcify and using the scripts to create a proposal encoding data for that contract, in the UI it should then show the data we gathered from the ABI

rossneilson:
https://docs.sourcify.dev/docs/api/

PR #879, opened 2022-05-17
feat: added sourcify with tests

0xvangrim:
# Description

Issue: https://github.com/DXgovernance/dxvote/issues/755

Adding a Sourcify API call to see if the contract chosen is verified or not. If it is chosen, we will decode the ABI and give the user an interface.

image

Difficult to test, but I added tests and tested it so it doesn't break any of the existing logic we use to show the contract interface.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [X ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

Manually and with unit tests.

# Checklist:

- [X ] My code follows the existing style of this project
- [X ] I have performed a self-review of my own code
- [ X] I have commented my code, particularly in hard-to-understand areas
- [ X] I have added tests that prove my fix is effective or that my feature works
- [X ] New and existing unit tests pass locally with my changes
- [ ] Any UI changes have been tested and made responsive for mobile views

### DIFF
--- a/src/hooks/Guilds/contracts/useDecodedCall.ts
+++ b/src/hooks/Guilds/contracts/useDecodedCall.ts
@@ -165,5 +165,5 @@ export const useDecodedCall = (call: Call) => {
     };
   }, [call, contracts, chainId]);
 
-  return decodedCall || { decodedCall: null, contract: null };
+  return decodedCall || { decodedCall: null, contract: null, approval: null};
 };
--- a/src/hooks/Guilds/contracts/useDecodedCall.ts
+++ b/src/hooks/Guilds/contracts/useDecodedCall.ts
@@ -12,6 +12,8 @@ import {
   SupportedAction,
 } from 'old-components/Guilds/ActionsBuilder/types';
 import { ERC20_APPROVE_SIGNATURE, ERC20_TRANSFER_SIGNATURE } from 'utils';
+import { useEffect, useState } from "react";
+import { lookUpContractWithSourcify } from 'utils/sourcify';
 
 const knownSigHashes: Record<string, { callType: SupportedAction; ABI: any }> =
   {
@@ -82,7 +84,7 @@ const getContractFromKnownSighashes = (data: string) => {
   };
 };
 
-export const decodeCall = (
+export const decodeCall = async (
   call: Call,
   contracts: RichContractData[],
   chainId: number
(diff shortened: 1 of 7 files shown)


src/hooks/Guilds/contracts/useDecodedCall.ts
```diff
@@ -124,21 +131,39 @@ export const bulkDecodeCallsFromOptions = (
...
+      );
+    } else {
+      setDecodedCall(null);
+    }
+    return () => {
+      isCancelled.current = true;
+    };
+  }, [call, contracts, chainId]);
+
+  return decodedCall || { decodedCall: null, contract: null };
```
MiltonTulli: missing approval in the fallback object.

src/utils/sourcify.ts
```diff
@@ -0,0 +1,10 @@
+export const lookUpContractWithSourcify = async ({ chainId, address }) => {
+  const baseUrl = `https://sourcify.dev/server/files/any`;
```
MiltonTulli, line 2: Do you think we store this in env/configs/constants file?

MiltonTulli (dismissed):

src/utils/sourcify.ts
0xvangrim: Sourcify will be used to lookup contracts that we are not storing ourselves. Or am I missing something in your question?

MiltonTulli: Just wondering if we should pull this url from config file, or env. Nothing serious

MiltonTulli (approved):

AugustoL (changes requested):
The only thing that It is missing is a test comparing a the abi and contact byecode hash from the richContractData with one from sourcify, I see you are only testing against a random contract/

0xvangrim:
> The only thing that It is missing is a test comparing a the abi and contact byecode hash from the richContractData with one from sourcify, I see you are only testing against a random contract/

Hm, I'm not certain why we would need to have that test? The `richContractData` and `sourcify` should never overlap. When the contract is not in the contract registry we will fetch it from sourcify, but only then. We will never make a double-check against both the registry and sourcify.

src/utils/sourcify.ts
0xvangrim: Ah I see, now I understand!

0xvangrim:
@AugustoL I've tested some of our `richContractData` that exists in our `dxvote`-file now and they haven't been been verified there yet.

AugustoL (approved):

AugustoL:
Approved, we see how we can add those tests for sourcify after then

PR #887, opened 2022-05-20
Feature/set permissions

### LINKED ISSUE #843: Set permissions core function call
We need an additional core action in the guild proposal actions builder. We currently have good examples for transfer and mint functions so this should follow the same pattern. We need a "Set Permissions" call. This allows our guilds to set their own specific permissions to control what is allowed and not allowed.
There is a lot we could do for this but for now, we will keep it simple. 

This is a call we do to the guild contract itself `setPermission(address[] memory asset,
        address[] to,
        bytes4[] functionSignature,
        uint256[] valueAllowed,
        bool[] allowance)`

We will show two main types of permissions to be set in the UI for now:
- Asset transfers
- Function calls

For asset calls we want an input for token and token amount (we already have these components in the UI). These will then encode as `asset` and `valueAllowed`. We should also add a button to set the token amount to max value (MAX_UINT). 
We also need a "To Address" input that is optional if left blank it will be ANY_ADDRESS. 
The remaining `functionSignature` should be set to`ANY_SIGNATURE`. 
We will also need to add special encoding conditions for transferring the native network token as this works differently. 

For function calls we want inputs for "To Address" and "Function signature". We will then need to hash the function signature to bytes4. I need to investigate more to find what exactly the other parameters should be set to for this permission type. 

ANY_ADDRESS = address(0xaAaAaAaaAaAaAaaAaAAAAAAAAaaaAaAaAaaAaaAa)
ANY_SIGNATURE = bytes4(0xaaaaaaaa)

We also need to display this when we show decoded proposal data (creating the proposal is encoding and we need to decode when showing existing proposal actions). We can just follow the same logic as transfer functions here and the code structure is set up to allow additional types. In future we will show more here but for now we can just display a simple bit of text describing the permissions being set.

dcrescimbeni:
# Description

Added "Set permissions" action.

Closes #843 

## Type of change

# How Has This Been Tested?

Added unit tests for the core functionality.

# Checklist:


dcrescimbeni:
Closed. Check: https://github.com/DXgovernance/dxvote/pull/889

PR #897, opened 2022-05-24
Feature/set permissions2

### LINKED ISSUE #843: Set permissions core function call

dcrescimbeni:
# Description

Implemented a new action: set permission.

Closes #843 

## Type of change

# How Has This Been Tested?

Added a test file to check the correct rendering and functionality of the component

# Checklist:


src/contexts/Guilds/guildAvailability.tsx
```diff
@@ -35,10 +35,10 @@ export const GuildAvailabilityContext =
   createContext<GuildAvailabilityContextInterface>({});
 
 const GuildAvailabilityProvider = ({ children }) => {
-  const routeMatch = useRouteMatch<{ guildId?: string }>(
-    '/:chainName/:guildId'
+  const routeMatch = useRouteMatch<{ guild_id?: string }>(
+    '/:chain_name/:guild_id'
```
rossneilson: Pretty sure this should be the previous `guildId` ?

src/old-components/Guilds/ActionsBuilder/CallDetails/index.tsx
```diff
@@ -54,6 +54,10 @@ const CallDetails: React.FC<ActionViewProps> = ({
     }
 
     if (type.startsWith('uint') || type.startsWith('int')) {
+      if (Array.isArray(value)) {
+        return <ParamDetail>{BigNumber.from(value[0]).toString()}</ParamDetail>;
```
rossneilson: We could probably just add this to a ternary isArray ? value[0] : value

src/old-components/Guilds/ActionsBuilder/SupportedActions/SetPermissions/AddressInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+const AddressInput: React.FC<AddressButtonProps> = ({
+  customToAddress,
+  anyAddressToggled,
+  validations,
+  destinationAvatarUrl,
+  parsedData,
+  handleCustomAddress,
+}) => {
+  return (
+    <Input
```
rossneilson: Would it be possible for us to reuse the existing address component here and just add on the "any" functionality?

src/old-components/Guilds/ActionsBuilder/SupportedActions/SetPermissions/SetPermissionsEditor.tsx
```diff
@@ -0,0 +1,313 @@
...
+        </TabButton>
+        <TabButton
+          aria-label="functions call tab"
+          active={activeTab === 1}
+          onClick={() => setActiveTab(1)}
+        >
+          Functions call
+        </TabButton>
+      </DetailWrapper>
+      {activeTab === 0 && (
```
rossneilson: Do you think it would be difficult to make the tab its own component that we can pass an array of objects to with the titles and children to render?

src/old-components/Guilds/ActionsBuilder/SupportedActions/SetPermissions/SetPermissionsInfoLine.tsx
```diff
@@ -0,0 +1,50 @@
+import { useMemo } from 'react';
+import { ActionViewProps } from '..';
+import { Segment } from '../common/infoLine';
+import Avatar from 'old-components/Guilds/Avatar';
+import { FiArrowRight, FiNavigation } from 'react-icons/fi';
+import useENSAvatar from 'hooks/Guilds/ether-swr/ens/useENSAvatar';
+import { MAINNET_ID, shortenAddress } from 'utils';
+import { ParsedDataInterface } from './types';
+
+// TODO: What is the most logical way to present the information? Is more information needed?
```
rossneilson, line 10: Nice todo, there's quite a lot we can do here and I'll make some follow up issues for it

src/old-components/Guilds/ActionsBuilder/CallDetails/index.tsx
rossneilson: Actually, we should instead make a loop. When in future we support arrays for permissions we will want to show all of them.
If it is an array you can loop over converting each to a string and keep the square brackets [123, 456]

src/old-components/Guilds/ActionsBuilder/SupportedActions/SetPermissions/AddressInput.tsx
dcrescimbeni: I don't seem to find an address input component, or maybe I'm not understanding the question

rossneilson: /src/old-components/Guilds/common/Form/AddressInput.tsx

dcrescimbeni: Done! It was a perfect fit.

I also added a "clear input" button to the component.

rossneilson (approved):

rossneilson:
Amazing work Dino, very impressive for one of your first issues!

PR #899, opened 2022-05-25
Feature/#855/header refactor

MiltonTulli:
# Description
Header refactor
Closes https://github.com/DXgovernance/dxvote/issues/855


src/locales/en/translation.json
```diff
@@ -4,9 +4,16 @@
     "addOption": "Add Option",
     "editOption": "Edit Option",
     "deleteOption": "Delete Option",
-    "optionLabel": "Option Label"
+    "optionLabel": "Option Label",
```
rossneilson: Not just something for this PR but we could quickly fix it here. I don't think we should be putting non guild specific text in guilds object, most of these will be reused outside of guilds

MiltonTulli: Agree. Not sure what is generic and what is guild specific. Hard one. Should I put all in the root obj? I think the only specific one for guilds is `create guild`

rossneilson: Yeah, safe bet to go for anything that says guild in it for now

MiltonTulli: Sure. Changed translation file and refactor. I think we can refactor later if we find some specific guild stuf, but for now almost all sounds like generic.

rossneilson (approved):

PR #936, opened 2022-11-09
Fix/add new contribution rewards

AugustoL:
# Description

Add new ContributionReward scheme into configs, this type of schemes was not designed to be detected automatically since it is a fork of legacy schemes, therefore it needs to be added manually to the config and we need to update the cache after adding it.


AugustoL:
The cache in xdai was updated and verified by @rossneilson

rossneilson (approved):

PR #942, opened 2022-11-24
Small refactor on ipfs proposal description hash

AugustoL:
- Supports future v1 ipfs hashes.
- Optimize a bit the description hash handling.

### DIFF
--- a/src/services/IPFSService.ts
+++ b/src/services/IPFSService.ts
@@ -49,15 +49,18 @@ export default class IPFSService {
   }
 
   async getContentFromIPFS(hash: string, timeout = 60000) {
-    const gatewayURLBaseList = [
-      'https://dxgov.mypinata.cloud/ipfs/',
-      'https://davi.mypinata.cloud/ipfs/',
-      'https://ipfs.io/ipfs/',
-      'https://gateway.ipfs.io/ipfs/',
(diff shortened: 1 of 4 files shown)


src/services/IPFSService.ts
```diff
@@ -49,15 +49,18 @@ export default class IPFSService {
...
-      'https://davi.mypinata.cloud/ipfs/',
-      'https://ipfs.io/ipfs/',
-      'https://gateway.ipfs.io/ipfs/',
-      'https://cloudflare-ipfs.com/ipfs/',
-      'https://dweb.link/ipfs/',
-      'https://infura-ipfs.io/ipfs/',
-    ];
+    const gatewayURLBaseList =
+      hash.substring(0, 2) == 'Qm'
+        ? [
```
rossneilson, line 54: These should support v1 I think

rossneilson:
This wont support the v1 already in mainnet? because its not working so testing is tough

src/services/IPFSService.ts
AugustoL: It does, is the other option, if it doesn't start with Qm it is v1.

AugustoL:
@rossneilson I cleared cache and only one (from two using v1 setting ens) proposal title was caught in the proposals page,

rossneilson (approved):

